PR #20, opened 2026-03-25
Add execute_command/assertion framework and migrate tests to runCommand format

eerxuan:
Add framework modules:
- framework/executor.py: execute_command, execute_admin_command
- framework/infra_exceptions.py: infrastructure error detection. Will update analyzer to use the same list of errors in following PR.
- framework/assertions.py: assertSuccess, assertFailure, assertResult

Migrate all existing tests from PyMongo driver calls to runCommand format using execute_command + framework assertions. This enables consistent error handling and result comparison across engines.

Update framework/__init__.py exports and simplify tests/__init__.py.

### DIFF
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -1,0 +1,61 @@
+name: PR Tests
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  test-mongodb:
+    name: Tests (MongoDB)
+    runs-on: ubuntu-latest
+
+    services:
+      mongodb:
+        image: mongo:8.2
+        ports:
+          - 27017:27017
+        options: >-
+          --health-cmd "mongosh --eval 'db.runCommand({ ping: 1 })'"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: pip install -r requirements.txt
+
+      - name: Run tests against MongoDB
+        run: |
+          pytest documentdb_tests/ \
+            --connection-string "mongodb://localhost:27017" \
+            --engine-name mongodb \
+            -n auto \
+            -v \
+            --json-report --json-report-file=${{ github.workspace }}/.test-results/mongodb-report.json \
+            --junitxml=${{ github.workspace }}/.test-results/mongodb-results.xml
+
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: test-results-mongodb
+          include-hidden-files: true
+          path: ${{ github.workspace }}/.test-results/
+          if-no-files-found: warn
+
+      - name: Generate test summary
+        if: always()
+        run: |
+          if [ -f ${{ github.workspace }}/.test-results/mongodb-report.json ]; then
+            python -m documentdb_tests.compatibility.result_analyzer -i ${{ github.workspace }}/.test-results/mongodb-report.json -o ${{ github.workspace }}/.test-results/mongodb-analysis.txt -f text || true
+            echo "## MongoDB Test Results" >> $GITHUB_STEP_SUMMARY
+            echo '```' >> $GITHUB_STEP_SUMMARY
+            cat ${{ github.workspace }}/.test-results/mongodb-analysis.txt >> $GITHUB_STEP_SUMMARY
+            echo '```' >> $GITHUB_STEP_SUMMARY
+          fi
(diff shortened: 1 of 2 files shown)


documentdb_tests/compatibility/tests/core/operator/stages/group/test_group_stage.py
```diff
@@ -6,123 +6,89 @@
...
 @pytest.mark.smoke
 def test_group_with_count(collection):
     """Test $group stage with count aggregation."""
-    # Arrange - Insert test data
     collection.insert_many([
-        {"name": "Alice", "department": "Engineering", "salary": 100000},
-        {"name": "Bob", "department": "Engineering", "salary": 90000},
-        {"name": "Charlie", "department": "Sales", "salary": 80000},
-        {"name": "David", "department": "Sales", "salary": 75000},
+        {"a": "A", "b": 100},
```
nitinahuja89: Why did we change the test data? I think the previous one is much more readable.

documentdb_tests/compatibility/tests/core/operator/stages/group/test_group_stage.py
```diff
@@ -6,123 +6,89 @@
...
-    result = list(collection.aggregate(pipeline))
-
-    # Assert - Verify results
-    assert len(result) == 2, "Expected 2 departments"
-
-    # Convert to dict for easier verification
-    dept_counts = {doc["_id"]: doc["count"] for doc in result}
-    assert dept_counts["Engineering"] == 2, "Expected 2 employees in Engineering"
-    assert dept_counts["Sales"] == 2, "Expected 2 employees in Sales"
+    result = execute_command(collection, {"aggregate": collection.name, "pipeline": [{"$group": {"_id": "$a", "count": {"$sum": 1}}}], "cursor": {}})
```
nitinahuja89: nit: for readability you might want to extract pipeline out into a variable above and then just use that here

documentdb_tests/compatibility/tests/core/operator/stages/group/test_group_stage.py
eerxuan: Keeping minimal test data intentionally — the spec here is "group by string field and count," so "A" is the
minimal case for a string. Application data like "Engineering" doesn't add testing value and introduces ambiguity
about whether string content/length matters. It also helps repo-wide consistency — every test uses the same
minimal patterns ("A", "B", 100) instead of each test inventing its own domain data.

documentdb_tests/compatibility/tests/core/collections/capped_collections/test_capped_collections_create.py
```diff
@@ -1,43 +1,20 @@
...
-        assert collection_info.get("capped") is True, "Collection should be capped"
-        
-        # Cleanup
-        database_client.drop_collection(collection_name)
-        
-    except OperationFailure as e:
-        raise
+def test_create_capped_collection(collection):
+    """Test creating a capped collection."""
+    db = collection.database
```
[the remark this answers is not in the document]
eerxuan: Will remove the drop, and append the unique name for the new collection to ensure uniqueness in parallel test runs. Just thinking when we test database creation, we cannot avoid the drop database in test case, becasue the fixture won't drop them.

```
def test_create_capped_collection(collection):
    """Test creating a capped collection."""
         result = execute_command(collection, {"create": collection.name + "_capped", "capped": True, "size": 100000})
         assertSuccess(result, {"ok": 1.0}, "Should create capped collection", raw_res=True, transform=lambda r: {"ok": r["ok"]})
```

documentdb_tests/framework/assertions.py
```diff
@@ -1,103 +1,130 @@
...
+    """Format a non-infra exception result into an assertion error message."""
+    code = getattr(result, 'code', None)
+    msg = getattr(result, 'details', {}).get('errmsg', str(result))
+    return f"[UNEXPECTED_ERROR] Expected success but got exception:\n{pprint.pformat({'code': code, 'msg': msg}, width=100)}\n"
 
 
-def assert_documents_match(
-    actual: List[Dict], expected: List[Dict], ignore_id: bool = True, ignore_order: bool = False
-):
+def assertSuccess(result: Union[Any, Exception], expected: Any, msg: str = None, raw_res: bool = False, transform: Optional[Callable] = None, ignore_order: bool = False):
```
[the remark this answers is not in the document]
eerxuan: transform is the internal extension point for assertSuccess/assertFailure — the two core assertion functions
that handle shared logic (infra error detection, formatting, logging). Tests can use transform directly for custom use.
Or, helpers like assertSuccessPartial wrap the core functions with a transform. This way we avoid duplicating
shared logic across assertion functions, and can add more helpers in the future (e.g., sort before compare for set operators) that
all go through the same core path. Updated the capped test to use assertSuccessPartial.

nitinahuja89 (approved):

PR #54, opened 2026-04-07
[RFC] Shared test constants and helpers for argument handling coverage

eerxuan:
This PR showcases a new test format on $divide arugment handling tests that reduces boilerplate code across all features.
- Refactor divide tests: split into 3 files, add type validation, argument handling, edge cases
- Reduce test boilerplate: shared constants, parametrize helpers, ErrorCode type
- Add expression type testing: expression_type_cases helper, inline setup/execute/assert

Pros:
- Consistent coverage — every operator tests the same cases, can't miss one
- Compact — with_expected(BSON_TYPE_SAMPLES, default=ERR, overrides={"null": None}) vs 15 lines
- Easy to scan — you see the file and immediately know what's covered
- Adding a new case (e.g., new BSON type) propagates to all operators automatically

Cons:
- Traceability — a failing test like test_divide_dividend_type[timestamp] requires you to:
  1. Find the test function
  2. See it uses BSON_TYPE_SAMPLES 
  3. Go to test_constants.py to see what value timestamp is
  4. Go to parametrize.py to understand how with_expected works
  - Before: everything was in one place — value, expression, expected, all visible
- Debugging is harder — the actual test input is assembled across multiple files
- New contributors need to learn the abstraction layer before they can write/read tests
- If we want to apply simiar pattern to other categories, we may need to design new helpers for different features. Design heavy in implementation.
- If the shared constant has a bug (wrong representative value), it silently affects all operators


eerxuan:
Discussed and decided to only apply the improvement to argument tests to avoid creating complicate abstractions. New PR: https://github.com/documentdb/functional-tests/pull/152

PR #55, opened 2026-04-07
Use strict numeric equality for test assertions

danielfrankcom:
This change replaces `==` with `_strict_equal` in the assertion helpers to catch two classes of mismatches:

- Float zero sign: `==` treats `-0.0` and `0.0` as equal per IEEE 754, but the sign is preserved through arithmetic and `$toString`. Tests that specify a signed zero expectation now actually enforce it.
- BSON numeric types: `==` treats `int` and `Int64` (and other numeric cross-type pairs) as equal, but they are distinct BSON types. Test expectations must now use the exact type returned by the server.

I also fixed the `test_smoke_expression_divide` expectations to use `float` literals since `$divide` returns doubles.


danielfrankcom:
Force push signs commits.

eerxuan (approved):

nitinahuja89 (approved):

Issue #34, opened 2026-04-03
Add compatibility test for stage operators (31 total)


eerxuan:
core	operator	stages	-	-	addFields
core	operator	stages	-	-	bucket
core	operator	stages	-	-	collStats
core	operator	stages	-	-	count
core	operator	stages	-	-	geoNear
core	operator	stages	-	-	group
core	operator	stages	-	-	indexStats
core	operator	stages	-	-	limit
core	operator	stages	-	-	listClusterCatalog
core	operator	stages	-	-	listSampledQueries
core	operator	stages	-	-	lookup
core	operator	stages	-	-	match
core	operator	stages	-	-	merge
core	operator	stages	-	-	out
core	operator	stages	-	-	planCacheStats
core	operator	stages	-	-	project
core	operator	stages	-	-	querySettings
core	operator	stages	-	-	queryStats
core	operator	stages	-	-	rankFusion
core	operator	stages	-	-	redact
core	operator	stages	-	-	replaceRoot
core	operator	stages	-	-	replaceWith
core	operator	stages	-	-	sample
core	operator	stages	-	-	search
core	operator	stages	-	-	searchMeta
core	operator	stages	-	-	set
core	operator	stages	-	-	skip
core	operator	stages	-	-	sort
core	operator	stages	-	-	unset
core	operator	stages	-	-	unwind
core	operator	stages	-	-	vectorSearch

PR #66, opened 2026-04-08
Stop wrapping expected value in assertResult

danielfrankcom:
The `assertResult` method is currently overfit for operator tests, making it unusable for other tests like aggregation pipelines.

This change moves the fixed `"result: expected"` structure to the call site, to allow the `assertResult` method to be reused more generally. If we decide we want to avoid this at the call site in the future, we can always add an operator-specific wrapper method which delegates to `assertResult` but I haven't included that here since there are relatively few calls to this method at the moment.

Relates to #34

### DIFF
--- a/documentdb_tests/compatibility/tests/core/operator/expressions/arithmetic/divide/test_operator_divide.py
+++ b/documentdb_tests/compatibility/tests/core/operator/expressions/arithmetic/divide/test_operator_divide.py
@@ -6,10 +6,10 @@ import pytest
 from bson import Decimal128, Int64
 
 from documentdb_tests.compatibility.tests.core.operator.expressions.utils.utils import (
+    assert_expression_result,
     execute_expression,
     execute_expression_with_insert,
 )
-from documentdb_tests.framework.assertions import assertResult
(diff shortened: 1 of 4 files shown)


eerxuan (approved):
I agree

documentdb_tests/compatibility/tests/core/operator/expressions/arithmetic/divide/test_operator_divide.py
```diff
@@ -707,7 +707,12 @@ class DivideTest(BaseTestCase):
 def test_divide_literal(collection, test):
     """Test $divide from literals"""
     result = execute_expression(collection, {"$divide": [test.dividend, test.divisor]})
-    assertResult(result, expected=test.expected, error_code=test.error_code, msg=test.msg)
+    assertResult(
```
nitinahuja89: let's add a operator specific wrapper in this case

danielfrankcom: Added a wrapper in `documentdb_tests/compatibility/tests/core/operator/expressions/utils/utils.py` since it's specific to expressions and shouldn't pollute the top-level file.

nitinahuja89 (approved):

PR #68, opened 2026-04-08
Add $match stage tests

danielfrankcom:
This change adds test cases for the `$match` aggregation stage, organized into files by property group:

- `test_match_core.py`: null/missing field matching, core equality/ordering, predicate semantics, empty predicates, large predicates
- `test_match_query_operators.py`: one representative per query operator category (comparison, logical, element, evaluation, array, bitwise, geospatial)
- `test_match_stage_position.py`: pipeline position behavior and `$text` first-stage
- `test_match_errors.py`: argument validation (non-object types) and restricted operators

Please note the tests are not intended to exhaustively cover things like BSON equality or other properties that would be thoroughly tested as part of separate query testing.

I also replaced `test_match_stage.py` with `test_smoke_match.py` (I think this overlaps with #64) since all of the existing cases were duplicated with the expanded coverage.

As part of this change, I've introduced a `StageTestCase` dataclass for pipeline stage tests, which should be reusable for other aggregation stage tests.

Depends on ~~#65, #66, #67, #73~~. Until those changes are merged, I expect any tests run on this branch to fail.

Ref: #34


documentdb_tests/compatibility/tests/core/operator/stages/match/test_match_core.py
```diff
@@ -0,0 +1,241 @@
...
+    StageTestCase,
+)
+from documentdb_tests.framework.assertions import assertResult
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+
+# Property [Null and Missing Field Matching]: {field: null} matches both
+# documents where the field is null-valued and documents where the field is
+# missing entirely.
+MATCH_NULL_MISSING_TESTS: list[StageTestCase] = [
```
eerxuan: Since find filter will behave the same in 
- find — filter
- update — q (query selector)
- delete — q
- findAndModify — query
- count — query
- distinct — query
- $match — aggregation stage
- bulkWrite — filter in each operation

We can:
A. Define the query predicate test cases once in a shared location (e.g., utils/query_test_cases.py), then have all feature tests parametrize over them with their own execution logic:
```
python
# stages/utils/query_test_cases.py
QUERY_NULL_MISSING_TESTS = [...]

# stages/match/test_match_core.py
@pytest.mark.parametrize("test_case", pytest_params(QUERY_NULL_MISSING_TESTS))
def test_match_null_missing(collection, test_case):
    # execute via aggregate pipeline

# find/test_find_filter.py
@pytest.mark.parametrize("test_case", pytest_params(QUERY_NULL_MISSING_TESTS))
def test_find_null_missing(collection, test_case):
    # execute via find command
```
B. Just run them under one place (tests/core/query-and-write/), assume very engine should use the same query path under each container.

documentdb_tests/compatibility/tests/core/operator/stages/utils/stage_test_case.py
```diff
@@ -0,0 +1,17 @@
...
+
+from documentdb_tests.framework.test_case import BaseTestCase
+
+
+@dataclass(frozen=True)
+class StageTestCase(BaseTestCase):
+    """Test case for pipeline stage tests."""
+
+    docs: Optional[list[dict[str, Any]]] = None
+    pipeline: Optional[list[dict[str, Any]]] = None
```
eerxuan: This is not optional.

eerxuan (commented):
The coverage is overall pretty good. We need to think about how to prevent duplicate testing.

documentdb_tests/compatibility/tests/core/operator/stages/utils/stage_test_case.py
eerxuan: okay, we can discuss if need to limit on 3.9 support. I feel we don't need as we don't have many dependencies to other legacy packages. Take from AI seems good, we get new features, and 3.9 is EOL.

```
Dropping 3.9 support seems reasonable for a test framework — it's not a library with downstream consumers. Python 3.9 hit EOL in October 2025, so it's already unsupported upstream. And the
test framework's dependencies (pytest, pymongo, etc.) are all fine on 3.10+.

If you bump to 3.10 minimum, you also get:
- kw_only=True on dataclasses (this issue)
- X | Y union syntax instead of Optional[X]
- match statement (not critical but nice)
- ParamSpec and TypeAlias in typing
```

danielfrankcom: Yes I agree, dropping 3.9 support given the EoL seems like the best approach here. Will not worry about this in the context of this PR though as it will be a larger change.

documentdb_tests/compatibility/tests/core/operator/stages/match/test_match_core.py
danielfrankcom: I agree the query predicate semantics are shared across those contexts, and duplicating coverage is worth avoiding.

That being said, the challenge with extracting the cases at the moment is that each context returns a different result shape. `match` and `find` return matched documents, `count` returns a number, `distinct` returns unique values, and `update`/`delete` return modification counts. The `expected` value in a shared test case can't be reused across these without deriving expectations programmatically, which makes the test cases less obvious to read where they are defined.

`$match` is the only consumer that currently exists, and as far as I can see atm is the only consumer that will be in the `stages` directory. I think designing the sharing abstraction with just one example to draw from may be premature, since there's no second consumer to validate the approach against. I'd rather keep the predicate cases here for now and refactor when the next consumer is added, since at that point there will be two real examples to inform the right shape and location for the shared files.

Not sure if you had something specific in mind for the structure of `QUERY_NULL_MISSING_TESTS`, but any version I experimented with ended up getting into the weeds with the way `expected` was defined.

documentdb_tests/compatibility/tests/core/operator/stages/test_stages_position_match.py
```diff
@@ -0,0 +1,180 @@
...
+    StageTestCase,
+)
+from documentdb_tests.framework.assertions import assertResult
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+
+# Property [Pipeline Position]: $match filters correctly regardless of its
+# position in the pipeline and composes with preceding stages that reshape
+# documents.
+MATCH_PIPELINE_POSITION_TESTS: list[StageTestCase] = [
```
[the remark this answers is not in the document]
danielfrankcom: `FOLDER_STRUCTURE.md` says "Integration tests (multiple same-level features interacting) → parent folders". I can see how these tests fit that pattern since they use `$match` alongside `$group`, `$project`, `$replaceRoot`, etc.

I think maybe this is a conceptual misunderstanding for me. Is the intent of the `FOLDER_STRUCTURE.md` rule about test isolation, so that failures under `match/` stay attributable to `$match` alone? Or is it about the conceptual scope of what `stages/` should own?

If it's the former I understand the concern, and can move these out to a dedicated file like `test_match_stage_position.py`, but the tests would still fail on a `$group` bug regardless of where they live in the tree.

If it's the latter then to me, the tests we're discussing here aren't verifying that stages compose correctly in a pipeline, they're verifying that `$match` produces the right output when its input documents have different shapes, which is `$match`-specific behavior that happens to require a preceding stage to set up. I think dedicated tests for stage composition at the `stages/` level make sense, things like verifying that output from one stage feeds correctly into the next as a general pipeline concern, but these tests we're discussing are not those.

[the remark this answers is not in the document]
eerxuan: > Is the intent of the FOLDER_STRUCTURE.md rule about test isolation, so that failures under match/ stay attributable to $match alone?

Yes this is a major intention. The generated report can have a clean view for leaf features, we know if it has failure, pretty confident the feature itself has issue not other dependency features.

> they're verifying that $match produces the right output when its input documents have different shapes

If testing this spec, we should just pass the "different shapes" directly into $match. But the tests here are using other stages to generate them in pipeline, it is integration test testing using $match in pipeline. Consider pipeline as a higher level(container) feature, similarly we considered expression as higher level(container) feature for expression operators.

documentdb_tests/compatibility/tests/core/operator/stages/match/test_match_core.py
eerxuan: Good point, there are more complexities. I guess we don't want to block this PR on this issue, , just raising this question so we can have consistent guideline onward. We can spend time refactoring it when we actually implement one of those features.

documentdb_tests/compatibility/tests/core/operator/stages/match/test_match_query_operators.py
```diff
@@ -0,0 +1,130 @@
...
+from documentdb_tests.compatibility.tests.core.operator.stages.utils.stage_test_case import (
+    StageTestCase,
+)
+from documentdb_tests.framework.assertions import assertResult
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+
+# Property [Query Operator Categories]: one representative per query operator
+# category functions correctly inside $match as a container.
+MATCH_QUERY_OPERATOR_TESTS: list[StageTestCase] = [
```
manojnks, line 16: cases like $lt , $in and more operator are missing since at least one of each should be added right?

eerxuan (approved):

eerxuan (changes requested):
Please fix CI failures.

danielfrankcom:
> Please fix CI failures.

No fixes needed, just waiting on #73

documentdb_tests/compatibility/tests/core/operator/stages/match/test_match_query_operators.py
danielfrankcom: I chose to use representative operators per category here rather than trying to enumerate every possible value, as I felt that it provided sufficient coverage given there are other tests for each of the operators.

This seems like a policy decision though, one that will apply to all aggregation pipeline tests.

manojnks: is there a place where we are testing $match with operators that are not here.

danielfrankcom: No, I don't think so, will add these to confirm each operator works in the `$match` container.

manojnks (approved):

manojnks:
Please fix CI failures.

danielfrankcom:
> Please fix CI failures.

Now that #73 is merged, I merged from upstream and the failures are resolved.

manojnks (approved):

eerxuan (approved):

PR #72, opened 2026-04-09
Add 92 smoke tests, no_parallel marker, and two-phase parallel handling: window operators, query-and-write, query-planning, sessions, administration, diagnostic, security (auth/roles/users)

fanyangv:
Closes https://github.com/documentdb/functional-tests/issues/60

- 92 smoke tests for: window operators (28), query-and-write commands (6), query planning (4), sessions (5), system administration (13), diagnostics (18), security (18)
- Add no_parallel pytest marker for 14 tests that modify global server state (kill sessions/ops, set parameters, drop all users/roles)
- Add pytest_collection_modifyitems hook to auto-deselect no_parallel tests under xdist
- Add pytest_sessionfinish hook to automatically run no_parallel tests sequentially after parallel phase
- Add worker_id fallback fixture for running without xdist
- Update README with two-phase execution docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,13 @@ pytest -n auto
 pytest -n auto -m smoke --connection-string mongodb://localhost:27017 --engine-name documentdb
 ```
 
+**Two-Phase Execution:**
+Some tests (e.g., `killAllSessions`, `setParameter`, `fsync`) modify global server state and cannot safely run in parallel. These are marked with `no_parallel`. When you use `-n`, the framework automatically:
+1. **Phase 1**: Runs all parallel-safe tests with multiple workers
+2. **Phase 2**: Runs `no_parallel` tests sequentially after Phase 1 completes
+
+This is handled transparently — no extra flags needed.
+
 **Performance Benefits:**
 - Significantly faster test execution with multiple workers
 - Scales with number of available CPU cores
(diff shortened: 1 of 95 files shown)


eerxuan:
**Verified locally: two-phase parallel handling works correctly**

- Phase 1 (parallel, `-n auto`): 674 tests passed in 20.75s — `no_parallel` tests deselected
- Phase 2 (sequential, automatic): 14 `no_parallel` tests passed in 5.12s

**One issue found:**

```
⚠️ Failed to merge JSON reports: [Errno 2] No such file or directory: '.report.json'
```

When `--json-report` is not explicitly passed, the `json_report_file` option still has a default value (`.report.json`) from the plugin. Phase 2 sees this default, generates a temp report, and tries to merge into a file that Phase 1 never wrote.

Fix: check if Phase 1 report actually exists before merging. If it doesn't exist and Phase 2's does, either copy Phase 2's report as the final report or skip silently (no warning):

```python
if phase1_json and phase2_json and os.path.exists(phase2_json):
    if os.path.exists(phase1_json):
        _merge_json_reports(phase1_json, phase2_json)
    else:
        # Phase 1 didn't write a report (--json-report not passed)
        # Skip merge silently — Phase 2 temp file will be cleaned up
        pass
```

Same applies to JUnit XML merge.

eerxuan:
Minor: `tempfile.mktemp` is deprecated (https://docs.python.org/3/library/tempfile.html#tempfile.mktemp). Consider `tempfile.NamedTemporaryFile(delete=False, suffix=".json", prefix="phase2_")` instead. Not a practical concern here, just a lint/best-practice thing.

fanyangv:
Addressed both review comments:

1. **Fixed report merge error when `--json-report` is not passed**: Now checks `os.path.exists(phase1_json)` before setting up Phase 2 report files, so we only generate/merge reports when Phase 1 actually wrote them. Same guard added for JUnit XML.

2. **Replaced deprecated `tempfile.mktemp`** with `tempfile.NamedTemporaryFile(delete=False, ...)` per Python docs recommendation.

Also rebased onto latest `main` and resolved a conflict in `pytest.ini` (kept both `engine_xfail` and `no_parallel` markers).

All CI checks pass locally: lint, unit tests, and full compatibility suite (9551 passed, 14 no_parallel passed).

eerxuan (approved):

Issue #10, opened 2026-03-24
Add compatibility test for string expression operators (20 total)


eerxuan:
core	operator	expressions	string	-	concat
core	operator	expressions	string	-	indexOfBytes
core	operator	expressions	string	-	indexOfCP
core	operator	expressions	string	-	ltrim
core	operator	expressions	string	-	regexFind
core	operator	expressions	string	-	regexFindAll
core	operator	expressions	string	-	regexMatch
core	operator	expressions	string	-	replaceAll
core	operator	expressions	string	-	replaceOne
core	operator	expressions	string	-	rtrim
core	operator	expressions	string	-	split
core	operator	expressions	string	-	strcasecmp
core	operator	expressions	string	-	strLenBytes
core	operator	expressions	string	-	strLenCP
core	operator	expressions	string	-	substr
core	operator	expressions	string	-	substrBytes
core	operator	expressions	string	-	substrCP
core	operator	expressions	string	-	toLower
core	operator	expressions	string	-	toUpper
core	operator	expressions	string	-	trim

PR #79, opened 2026-04-09
Add string operator $replaceAll/$replaceOne tests

eerxuan:
Add compatibility tests for the $replaceAll and $replaceOne operators.

Ref: #10

Tests cover:
- Core replacement behavior
- Encoding and special characters
- Type errors and invalid arguments
- Null/missing handling
- Size limits
- Invariants and type precedence
- Whitespace handling
- Expression arguments and document field usage

497 tests total (3 known failures: empty_find_multibyte tests fail due to pymongo UTF-8 decoding of invalid byte sequences).


danielfrankcom (commented):
Seems like we might need to add some kind of pytest `mark` to indicate these tests are expected to fail for MongoDB as previously discussed? If we override the merge conditions here we'll end up with all future PRs failing these checks which seems undesirable.

danielfrankcom (commented):
Won't the use of `xfail` mean that the compatibility tests expect **all** compatible databases to fail this test?

Since in this case it may be a database-specific issue, we may need a different approach here or we may cause this to fail for other compatible implementations even though the same problem doesn't exist.

danielfrankcom:
> Won't the use of `xfail` mean that the compatibility tests expect **all** compatible databases to fail this test?
> 
> Since in this case it may be a database-specific issue, we may need a different approach here or we may cause this to fail for other compatible implementations even though the same problem doesn't exist.

I'm not sure whether this would work for us in practice, but it looks like the [`xfail` docs](https://docs.pytest.org/en/latest/reference/reference.html#pytest.mark.xfail) show a `condition` parameter. We may be able to isolate the `xfail` behavior only to MongoDB.

danielfrankcom:
Seems like the `condition` needs to evaluate at setup time so accessing the engine name would be tricky.

Haven't tested this, but we might need to do something like this:

```python
def pytest_configure(config):
    # We might be able to do this in the static `.ini` file instead, not sure.
    config.addinivalue_line(
        "markers",
        "engine_xfail(engine, reason): expected failure for a specific engine",
    )


def pytest_runtest_setup(item):
    for marker in item.iter_markers("engine_xfail"):
        if item.config.engine_name == marker.kwargs.get("engine"):
            pytest.xfail(marker.kwargs.get("reason", ""))
```

And then in the test file:

```python
marks=(pytest.mark.engine_xfail(engine="mongodb", reason="MongoDB returns invalid UTF-8 for empty-find multibyte"),),
```

documentdb_tests/conftest.py
```diff
@@ -45,6 +45,10 @@ def pytest_addoption(parser):
 
 def pytest_configure(config):
     """Configure pytest with custom settings."""
+    config.addinivalue_line(
```
danielfrankcom: I think we can just add this in `pyproject.toml` since the other marks are defined there:

```
markers = [
    # ... existing markers ...
    "engine_xfail(engine, reason): expected failure for a specific engine",
]
```

documentdb_tests/framework/assertions.py
```diff
@@ -12,6 +12,17 @@
 
 from documentdb_tests.framework.infra_exceptions import INFRA_EXCEPTION_TYPES as _INFRA_TYPES
 
+_MAX_REPR_LEN = 1000
```
danielfrankcom, line 15: What is the intent with this change? Maybe I don't follow the reasoning, but from a failure like [this one](https://github.com/documentdb/functional-tests/actions/runs/24373391627/job/71181626156) which timed out, none of the individual failures are that long:

```

________________ test_replaceall_core_cases[edge_find_at_start] ________________
[gw1] linux -- Python 3.12.13 /opt/hostedtoolcache/Python/3.12.13/x64/bin/python
documentdb_tests/compatibility/tests/core/operator/expressions/string/replaceAll/test_replaceAll_core.py:457: in test_replaceall_core_cases
    assertResult(
documentdb_tests/framework/assertions.py:214: in assertResult
    assertSuccess(result, expected, msg)
documentdb_tests/framework/assertions.py:114: in assertSuccess
    assert _strict_equal(result, expected), error_text
E   AssertionError: [RESULT_MISMATCH] $replaceAll edge: find at start
E     
E     Expected:
E     'Xdef'
E     
E     Actual:
E     [{'result': 'Xdef'}]
E     
E   assert False
E    +  where False = _strict_equal([{'result': 'Xdef'}], 'Xdef')

```

I'm not sure why it's so slow (I assume it does have to do with the log output but not certain), but truncating each test case doesn't seem like it would fix the problem.

I'd also be concerned about losing real information in the case of a valuable failure, especially an intermittent one which may be hard to reproduce if we don't have the information from the first time it fails.

Maybe we can leave this change out of this PR, and if we think it's valuable spin it out as a separate change where we can test it for these concerns?

documentdb_tests/compatibility/tests/core/operator/expressions/string/replaceAll/test_replaceAll_core.py
```diff
@@ -0,0 +1,499 @@
...
+    ReplaceAllTest(
+        "empty_find_multibyte_2byte",
+        input="\u00e9",
+        find="",
+        replacement="X",
+        expected="X\u00e9X",
+        msg="$replaceAll empty string: find multibyte 2byte",
+        marks=(
+            pytest.mark.engine_xfail(
+                engine="mongodb",
```
danielfrankcom, line 339: I'm not sure the specific error that occurs in these cases, but it seems like `xfail` has a `raises=` field that we can use to specify the exception to expect.

This seems more precise, and will help avoid a situation where the test starts failing because it was broken, rather than failing for the reason that is intended.

documentdb_tests/compatibility/tests/core/operator/expressions/string/replaceAll/test_replaceAll_core.py
```diff
@@ -0,0 +1,499 @@
...
+                reason="MongoDB returns invalid UTF-8 for empty-find multibyte (#95)",
```
danielfrankcom, line 340: It looks like there is a `strict=` property which is `False` by default, which means if the test passes the test suite will not consider this a problem.

Rather than specifying `strict=True` on every invocation of this mark, I think we should set [`strict_xfail`](https://docs.pytest.org/en/stable/reference/reference.html#confval-strict_xfail) in the pytest options.

documentdb_tests/conftest.py
```diff
@@ -58,6 +62,13 @@ def pytest_configure(config):
         config.connection_string = "mongodb://localhost:27017"
 
 
+def pytest_runtest_setup(item):
+    """Apply engine-specific xfail markers."""
+    for marker in item.iter_markers("engine_xfail"):
+        if getattr(item.config, "engine_name", None) == marker.kwargs.get("engine"):
+            pytest.xfail(marker.kwargs.get("reason", ""))
```
danielfrankcom: Can be done as a follow-up, but it seems like we will need to update the test results report format. It currently shows these as unknown tests which makes them seem like an unaccounted for problem:

```

FAILED TESTS
--------------------------------------------------------------------------------

  UNKNOWN (3):

    compatibility/tests/core/operator/expressions/string/replaceAll/test_replaceAll_core.py::test_replaceall_core_cases[empty_find_multibyte_2byte]
      Tags: 
      Duration: 0.00s
      Error: ...

    compatibility/tests/core/operator/expressions/string/replaceAll/test_replaceAll_core.py::test_replaceall_core_cases[empty_find_multibyte_3byte]
      Tags: 
      Duration: 0.00s
      Error: ...

    compatibility/tests/core/operator/expressions/string/replaceAll/test_replaceAll_core.py::test_replaceall_core_cases[empty_find_multibyte_4byte]
      Tags: 
      Duration: 0.00s
      Error: ...

================================================================================```
```

Unrelated but a few other things I noticed about this report:
- There is a trailing ``` at the end.
- The "RESULTS BY TAG" section shows all 100% despite the failures which is a bit confusing since at first glance it seems like that should cover all tests, but in practice I think some tests are just untagged

danielfrankcom (changes requested):
Most comments are around the introduction of `xfail`, which I guess is the downside of folding this change into an existing PR with test additions.

documentdb_tests/framework/assertions.py
eerxuan: It was slow because of the size limit tests, when they fail we have very long out put, the long string is printed 3 times in log for each test case. I feel it's rare an actual value will take that long given we are keeping test simple. I can expand the limit to 10000 though.

>   E AssertionError: [RESULT_MISMATCH] $replaceAll size limit: success input max
  E
  E Expected:
  E
  'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaa... (truncated, 16777217 chars total)
  E
  E Actual:
  E [{'result':
  'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa...
  (truncated, 16777231 chars total)
  E
  E assert False
  E + where False = _strict_equal([{'result':
  'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

danielfrankcom: Ah gotcha, that makes more sense. Did we get a run of this PR in its failing state before the compatibility fixes, but after this truncation to confirm it completed faster? If not maybe we can intentionally break it to confirm this fixes the issue but doesn't truncate anything we want to keep?

documentdb_tests/conftest.py
eerxuan: Added them.  The markers are defined in two places now .ini and .toml file, will need another PR to resolve this.

documentdb_tests/compatibility/tests/core/operator/expressions/string/replaceAll/test_replaceAll_core.py
eerxuan: Good point, seems strict= is also useful when mongodb fixes the issue, added both. Strict is applied for all xfail marker.

documentdb_tests/compatibility/tests/core/operator/expressions/string/replaceAll/test_replaceAll_core.py
eerxuan: Added.

documentdb_tests/conftest.py
eerxuan: Updated analyzer, now it handles xfail correctly and will error if see xpass, becuase we have strict, those should just have outcome FAILED.

documentdb_tests/conftest.py
danielfrankcom: Oh yeah, didn't notice that. When I was testing locally it wasn't picking up the `pytest.ini` file and causing warnings, maybe need to figure out which should remain and sync them, but agree this is a separate concern.

documentdb_tests/framework/assertions.py
eerxuan: Yes, added truncate before fixing test.

documentdb_tests/framework/assertions.py
```diff
@@ -126,14 +137,24 @@ def assertSuccess(
...
-        result = _sort_if_list(result)
-        expected = _sort_if_list(expected)
-
-    assert _strict_equal(result, expected), error_text
+    error_text += f"\n\nExpected:\n{_truncate_repr(expected)}"
+    error_text += f"\n\nActual:\n{_truncate_repr(result)}\n"
+
+    _large = len(repr(result)) > _MAX_REPR_LEN or len(repr(expected)) > _MAX_REPR_LEN
+
+    if ignore_doc_order and isinstance(result, list) and isinstance(expected, list):
```
danielfrankcom: Looks like the previous change got a bit mangled in the merge here. If we don't want to use `_sort_if_list` then we should remove the method too.

imo it's clearer to mutate the values with `_sort_if_list` since it helps avoid some of the nesting and the duplicate branch. Consider:

```python
    if ignore_doc_order:
        result = _sort_if_list(result)
        expected = _sort_if_list(expected)

    if _large:
        if not _strict_equal(result, expected):
            raise AssertionError(error_text)
    else:
        assert _strict_equal(result, expected), error_text
```

eerxuan: Fixed

danielfrankcom (approved):

danielfrankcom (approved):

nitinahuja89 (approved):

Issue #14, opened 2026-03-24
Add compatibility test for accumulator expression operators (7 total)


eerxuan:
core	operator	expressions	accumulator	-	avg
core	operator	expressions	accumulator	-	first
core	operator	expressions	accumulator	-	last
core	operator	expressions	accumulator	-	max
core	operator	expressions	accumulator	-	min
core	operator	expressions	accumulator	-	setUnion
core	operator	expressions	accumulator	-	sum

PR #88, opened 2026-04-11
Add $sum expression tests

danielfrankcom:
This change adds tests for the $sum expression.

Ref #14


documentdb_tests/framework/error_codes.py
```diff
@@ -4,11 +4,13 @@
 """
 
 DIVIDE_BY_ZERO_ERROR = 2
+FAILED_TO_PARSE_ERROR = 9
```
sathvikprasad97, line 7: Are we using this error case anywhere?

documentdb_tests/framework/error_codes.py
```diff
@@ -4,11 +4,13 @@
...
 TYPE_MISMATCH_ERROR = 14
 EXPRESSION_TYPE_MISMATCH_ERROR = 16020
 MODULO_ZERO_REMAINDER_ERROR = 16610
 MODULO_NON_NUMERIC_ERROR = 16611
 MORE_THAN_ONE_DATE_ERROR = 16612
+INVALID_DOLLAR_FIELD_PATH = 16872
```
sathvikprasad97, line 35: Are we using this error case anywhere?

sathvikprasad97 (changes requested):

documentdb_tests/framework/error_codes.py
danielfrankcom: Thanks, these were left over from rebasing.

documentdb_tests/framework/error_codes.py
danielfrankcom: Thanks, these were left over from rebasing.

sathvikprasad97 (approved):

eerxuan (changes requested):
Can you add Banker's rounding tests
— Only one rounding test (max_coeff + 1). Missing
  halfway-to-even tie-breaking cases per section 7 ("Rounding behavior: Precision
  loss scenarios")

eerxuan (approved):

PR #87, opened 2026-04-10
Added conditional $switch tests

vic-tsang:
- Added tests for $switch conditional expression

Ref: Issue #17

PR #83, opened 2026-04-10
added boolean $or test cases and ExpressionTestCase template

vic-tsang:
- Added tests for $or boolean expression 
- Added ExpressionTestCase template to be shared among expression tests 
- Added test_expressions_boolean_coercion.py for testing common boolean coercion 

Ref: Issue #17

Issue #17, opened 2026-03-24
Add compatibility test for remaining boolean, conditional, object, misc expression operators (14 total)


eerxuan:
core	operator	expressions	-	-	literal
core	operator	expressions	text	-	meta
core	operator	expressions	boolean	-	and
core	operator	expressions	boolean	-	not
core	operator	expressions	boolean	-	or
core	operator	expressions	conditional	-	cond
core	operator	expressions	conditional	-	ifNull
core	operator	expressions	conditional	-	switch
core	operator	expressions	misc	-	getField
core	operator	expressions	misc	-	rand
core	operator	expressions	object	-	mergeObjects
core	operator	expressions	object	-	setField
core	operator	expressions	object	-	unsetField
core	operator	expressions	variable	-	let

PR #92, opened 2026-04-13
Added tests for $cond

vic-tsang:
This PR contains

- Added tests for $cond 

Notes:
- Truthy/Falsy tests is tested comprehensively in test_expressions_boolean_coercion.py from #83 
- return types is packaged together with test_conditional_common.py from #87 

Ref: Issue #17 
Depends on #83 for ExpressionTestCase reuse


vic-tsang:
rebase files

documentdb_tests/compatibility/tests/core/operator/expressions/conditional/cond/test_cond_branching_and_correlation.py
```diff
@@ -0,0 +1,142 @@
...
+    ),
+    ExpressionTestCase(
+        "lazy_then_divide_by_zero_with_syntax_error",
+        expression={"$cond": {"if": "$flag", "then": {"$divide": ["$val"]}, "else": "$val"}},
+        doc={"flag": False, "val": 1, "zero": 0},
+        error_code=EXPRESSION_TYPE_MISMATCH_ERROR,
+        msg="Syntax errors are caught at parse time even in untaken branches",
+    ),
+    ExpressionTestCase(
+        "lazy_else_log_zero",
```
fanyangv: **Redundant lazy evaluation tests:** `lazy_else_divide_by_zero` and `lazy_else_log_zero` test the same thing — untaken else
  branch is not evaluated at runtime. The different error source (`$divide` vs `$log`) doesn't add meaningful coverage. Consider
  removing one.

fanyangv (commented):
1 — Invalid reference tests scope:

  **General comment (applies across operator PRs):** Tests like `invalid_field_path_dollar` (`"$"`) and `undefined_variable`
  (`"$$foo"`) validate the expression parser, not `$cond`-specific logic. The same errors would occur in any operator that accepts
  field paths. If we duplicate these in every operator PR, we'll end up with N copies of the same parser tests scattered across
  `$and`, `$or`, `$not`, `$cond`, etc. Consider moving these to a shared expression parsing test file (e.g.,
  `test_expression_parsing_errors.py`) that validates invalid field paths and undefined variables once, independent of any specific
  operator.

2 — Parametrize style:
**General comment (applies across all operator PRs):** All operator PRs should use `pytest_params()` from
  `documentdb_tests.framework.parametrize` instead of `ids=lambda t: t.id`.

Why: `BaseTestCase` has a `marks` field (`marks: tuple = field(default=())`) specifically designed to support per-test-case
  pytest marks like `pytest.mark.skip` or `pytest.mark.xfail`. The `ids=lambda` approach only extracts the test ID for display —
  pytest never sees the `marks` tuple, so any marks on individual test cases are silently ignored. `pytest_params()` wraps each case
  in `pytest.param(t, id=t.id, marks=t.marks)`, which passes both `id` and `marks` through to pytest correctly.

Current state across PRs:
- `$and` (#96): ✅ uses `pytest_params()` consistently
- `$or` (#83): ❌ uses `ids=lambda` in all 8 parametrize calls
- `$not` (#93): ❌ uses `ids=lambda` in 9 of 10 calls (combination file uses `pytest_params`)
- `$cond` (#92): ❌ uses `ids=lambda` in all 6 parametrize calls

Suggest aligning all PRs on `pytest_params()`.

vic-tsang:
pytest_params for other PRs has been fixed by @danielfrankcom in #114

documentdb_tests/compatibility/tests/core/operator/expressions/conditional/cond/test_cond_branching_and_correlation.py
vic-tsang: dropped lazy_else_log_zero test

vic-tsang:
Removed  invalid reference tests scope

fanyangv (approved):
NIT: The PR description "This PR fixes the following...", "fixes" seems misleading. Suggest to edit.

vic-tsang:
Updated PR descriptions

eerxuan (approved):

PR #93, opened 2026-04-13
Added tests for $not

vic-tsang:
This PR contains
- added tests for $not
- updated combination tests shared with $or test_expressions_combination_boolean.py

Ref: Issue #17 
Depends on ExpressionTestCase Template in #83


vic-tsang:
rebase files

fanyangv (changes requested):
1. **Combination file conflict:** Both this PR and the `$and` PR (#96) create `test_expressions_combination_boolean.py` with
  different content. The `$and` version covers `$and`/`$or`/`$not` combos (197 lines); this one covers `$or`/`$not` combos (163
  lines). Please consolidate into a single version in whichever PR lands first, and rebase the other on top.
2. **Missing return type test:** The `$or` PR (#83) includes a return type verification using `{"$type": {"$or": ...}}` →
  `"bool"` in `test_or_truthiness.py`. Please add an equivalent check for `$not` in `test_not_truthiness.py` (e.g., `{"$type":
  {"$not": [True]}}` → `"bool"`).
3. **Parametrize inconsistency:** Please use `pytest_params()` from `documentdb_tests.framework.parametrize` instead of
  `ids=lambda t: t.id` for all parametrize calls across the 3 boolean operator PRs. The lambda approach silently ignores the `marks`
  field on `BaseTestCase`, so per-test-case marks like `pytest.mark.skip` or `pytest.mark.xfail` won't take effect. `pytest_params()`
  handles both `id` and `marks` correctly. The `$and` PR (#96) already uses it consistently — let's align `$or` and `$not` to match.

vic-tsang:
- Rebase to fix combination file conflict 
- Fixed pytest_params
- Added return type tests

fanyangv (approved):

eerxuan (approved):

PR #96, opened 2026-04-14
Added test for boolean expression $and

vic-tsang:
This PR containts 
- Added tests for $and boolean expression
- updated combination tests shared with $and in test_expressions_combination_boolean.py

Ref: Issue #17 
Depends ExpressionTestCase template on #83


fanyangv (changes requested):
Comparing with the `$or` PR (#83), two things stand out:
1. **Missing truthiness tests.** The `$or` PR has a dedicated `test_or_truthiness.py` covering truthiness combos (all-falsy,
  all-numeric-zeros, mixed), order sensitivity (truthy at first/middle/last position), and return type verification via `$type`. The
  `$and` PR has no equivalent — please add a `test_and_truthiness.py` with analogous coverage.
3. **Inconsistent file structure.** The `$or` PR uses separate files per dimension (field_path, null_missing, nan_infinity,  expression_types). This PR consolidates them into fewer files. Consider aligning the structure so both operators follow the same
  pattern for easier navigation and maintenance.

vic-tsang:
Restructured $and file structure to use similar structure as $or

documentdb_tests/compatibility/tests/core/operator/expressions/boolean/and/test_and_expression_types.py
```diff
@@ -0,0 +1,40 @@
...
+        msg="Nested $and with mixed returns false",
+    ),
+    ExpressionTestCase(
+        "nested_and_inner_true",
+        expression={"$and": [{"$and": [True, True]}, True]},
+        expected=True,
+        msg="Nested $and with inner true should return true",
+    ),
+    ExpressionTestCase(
+        "nested_and_all_true",
```
fanyangv, line 28: nested_and_all_true is identical to nested_and_inner_true above (same expression and expected value). Consider differentiating it to test a distinct code path, e.g.:

  ExpressionTestCase(
      "nested_and_all_true",
      expression={"$and": [{"$and": [True, True]}, {"$and": [True, True]}]},
      expected=True,
      msg="Nested $and with all true should return true",
  ),

  This exercises multiple nested $and sub-expressions rather than duplicating the single-nested case.

fanyangv (changes requested):

documentdb_tests/compatibility/tests/core/operator/expressions/boolean/and/test_and_expression_types.py
vic-tsang: fixed

eerxuan (approved):

PR #98, opened 2026-04-14
Add $sort stage tests

danielfrankcom:
This change adds test cases for the `$sort` aggregation stage, organized into files by property group:

- `test_sort_basic_ordering.py`: ascending/descending ordering for all BSON types, equal-value tiebreaking
- `test_sort_numeric.py`: cross-type numeric interleaving (int32, int64, double, Decimal128)
- `test_sort_type_comparison.py`: BSON type comparison order, within-type ordering rules
- `test_sort_arrays.py`: array sort key extraction (min for asc, max for desc), empty/nested arrays
- `test_sort_key_resolution.py`: compound sort left-to-right precedence, null/missing field behavior
- `test_sort_field_paths.py`: dot notation traversal, array-of-objects extraction, field name acceptance
- `test_sort_direction_values.py`: accepted sort order values (int64, double truncation, Decimal128 rounding, `$meta`)
- `test_sort_order_errors.py`: sort order type errors, range errors, error precedence
- `test_sort_spec_errors.py`: non-object/empty spec validation, field path validation, compound key limit, duplicate keys
- `test_sort_meta_errors.py`: `$meta` non-meta object, extra keys, and invalid value errors
- `test_stages_position_sort.py`: pipeline position behavior, `$meta` sort with prerequisite stages (`$text`, `$geoNear`)

As part of this change I also:

- Introduced `build_raw_bson_doc` helper for constructing BSON documents with duplicate keys
- Added new error codes and test constants used by the sort tests

Please note this change introduces similar machinery to #68, and a rebase will be required once one or the other is merged.

Ref: #34


documentdb_tests/compatibility/tests/core/operator/stages/sort/test_sort_direction_values.py
```diff
@@ -0,0 +1,161 @@
...
+        docs=[{"_id": 1, "v": 20}, {"_id": 2, "v": 10}],
+        pipeline=[{"$sort": {"v": 1, "s": {"$meta": "searchScore"}}}],
+        expected=[{"_id": 2, "v": 10}, {"_id": 1, "v": 20}],
+        msg="$sort should accept {$meta: 'searchScore'} as a sort order value",
+    ),
+    StageTestCase(
+        "accept_meta_textscore",
+        docs=[{"_id": 1, "v": 20}, {"_id": 2, "v": 10}],
+        pipeline=[{"$sort": {"v": 1, "t": {"$meta": "textScore"}}}],
+        error_code=QUERY_METADATA_NOT_AVAILABLE_ERROR,
```
manojnks, line 112: message say it should accept but the case is for failure.
same for the next one as well

documentdb_tests/compatibility/tests/core/operator/stages/sort/test_sort_field_paths.py
```diff
@@ -0,0 +1,175 @@
...
+        docs=[
+            {"_id": 1, "a": {"b": 30}},
+            {"_id": 2, "a": {"b": 10}},
+        ],
+        pipeline=[{"$sort": {"a.b": 1}}],
+        expected=[
+            {"_id": 2, "a": {"b": 10}},
+            {"_id": 1, "a": {"b": 30}},
+        ],
+        msg="$sort should traverse embedded documents via dot notation",
```
manojnks: need to add dec of the same case

documentdb_tests/compatibility/tests/core/operator/stages/sort/test_sort_field_paths.py
```diff
@@ -0,0 +1,175 @@
...
+            {"_id": 1, "a": [{"b": 30}, {"b": 10}]},
+            {"_id": 2, "a": [{"b": 25}, {"b": 5}]},
+            {"_id": 3, "a": [{"b": 20}, {"b": 15}]},
+        ],
+        msg=(
+            "$sort descending should use max of nested field values"
+            " extracted from array of objects"
+        ),
+    ),
+    StageTestCase(
```
manojnks, line 80: add dec as well

documentdb_tests/compatibility/tests/core/operator/stages/sort/test_sort_key_resolution.py
```diff
@@ -0,0 +1,161 @@
...
+        pipeline=[{"$sort": {"a": 1, "b": 1}}],
+        expected=[
+            {"_id": 2, "a": 1, "b": 20},
+            {"_id": 4, "a": 1, "b": 40},
+            {"_id": 3, "a": 2, "b": 10},
+            {"_id": 1, "a": 2, "b": 30},
+        ],
+        msg="$sort should sort by first field then by second field left to right",
+    ),
+    StageTestCase(
```
manojnks, line 34: why does it matter

documentdb_tests/compatibility/tests/core/operator/stages/sort/test_sort_key_resolution.py
```diff
@@ -0,0 +1,161 @@
...
+        pipeline=[{"$sort": {"b": 1, "a": 1}}],
+        expected=[
+            {"_id": 3, "a": 2, "b": 10},
+            {"_id": 2, "a": 1, "b": 20},
+            {"_id": 1, "a": 2, "b": 30},
+            {"_id": 4, "a": 1, "b": 40},
+        ],
+        msg="$sort should use key insertion order, not alphabetical field name order",
+    ),
+    StageTestCase(
```
manojnks, line 51: also add case l for a:-1, b:1 and a:-1, b:-1

documentdb_tests/compatibility/tests/core/operator/stages/sort/test_sort_key_resolution.py
```diff
@@ -0,0 +1,161 @@
...
+            {"_id": 1, "v": None},
+            {"_id": 2},
+            {"_id": 3, "v": [None]},
+            {"_id": 4, "v": None},
+            {"_id": 5},
+            {"_id": 6, "v": [None]},
+        ],
+        msg="$sort should sort null, missing, and [null] equivalently, interleaved by _id",
+    ),
+    StageTestCase(
```
manojnks, line 128: this is same as `nested_200_component_path_accepted` can we remove this

documentdb_tests/compatibility/tests/core/operator/stages/test_stages_position_sort.py
```diff
@@ -0,0 +1,221 @@
...
+        ],
+        pipeline=[
+            {"$project": {"neg": {"$multiply": ["$v", -1]}}},
+            {"$sort": {"neg": 1}},
+            {"$project": {"_id": 1}},
+        ],
+        expected=[{"_id": 2}, {"_id": 3}, {"_id": 1}],
+        msg="$sort should sort on a field computed by a preceding $project",
+    ),
+    StageTestCase(
```
manojnks, line 122: can you add more combination of sort, project and group.
it would be good cases to have.
chimera has optimization like pulling sort into group.

documentdb_tests/framework/error_codes.py
```diff
@@ -4,11 +4,27 @@
 """
```
manojnks, line 4: can you also get approval for this by yunxuan

danielfrankcom: Which part are you referring to here?

documentdb_tests/compatibility/tests/core/operator/stages/sort/test_sort_key_resolution.py
danielfrankcom: Because the `$sort` key priority is supposed to be determined from left to right. If an implementation sorts the keys themselves, it would treat `a` as higher priority in this case, which would fail the test.

documentdb_tests/compatibility/tests/core/operator/stages/sort/test_sort_key_resolution.py
danielfrankcom: I think it's important to keep this one because it's the simplest version of the test case.

If the nonexistent field behavior is broken, this is the more obvious one to fail. If we only had the other one, it would seem like a failure relating to the 200 nested components, which would not be the cause.

The only reason the 200 component path is the same as this is because of the 180 vs 200 nesting limit mismatch in docs vs sort keys. Ideally they would support the same limit and it wouldn't have to be the same as this.

documentdb_tests/framework/error_codes.py
manojnks: error code file.
since this would framework changes

danielfrankcom: I don't think we've typically considered changes to `error_codes.py` that way, since a pretty good number of operators introduce at least 1 error code. They're pretty standardized and don't leave much room for interpretation with their location and format.

Happy to ping @eerxuan but probably not necessary in general.

danielfrankcom: Closing this as these were tackled as part of #107 to prevent conflicts, and they will be removed from this PR as part of an upcoming update.

documentdb_tests/compatibility/tests/core/operator/stages/test_stages_position_sort.py
danielfrankcom: Added these and a few other integration tests in a new file

danielfrankcom:
Due to the new integration test changes, this PR needs #68 to be merged first. It adds the `ignore_doc_order` parameter to `assertResult`, which the integration tests require.

danielfrankcom:
The #68 dependency was merged, so this should now be ready for final review.

documentdb_tests/framework/error_codes.py
manojnks: sounds good

manojnks (approved):

eerxuan (approved):

PR #102, opened 2026-04-15
Add $let and $literal tests

danielfrankcom:
This change adds tests for `$let` and `$literal` expressions. These tests were originally written by [@mitchell-elholm](https://github.com/mitchell-elholm).

**$let tests:**
- Core value passthrough for all BSON types, arithmetic, and string operations
- Variable scoping: nested `$let`, shadowing, variable swap, triple nesting
- Invalid argument validation: missing `vars`/`in`, non-object `vars`, unknown fields
- Null and missing field handling
- Path lookup through variables (`$$x.a.b`), including nested objects and arrays
- Variable naming rules: invalid start characters, invalid characters mid-name

**$literal tests:**
- BSON type passthrough including boundary values, special floats, and Decimal128 precision
- Expression suppression: `$add`, `$match`, `$group`, field paths, and system variables returned as-is
- Argument formats: nested objects, arrays, mixed special values
- `$project` disambiguation: `$literal` values vs inclusion/exclusion flags
- Special field names: dollar-prefixed, dot-containing, and empty field names

**Combination tests:**
- Nested `$let` scoping and shadowing across multiple levels
- `$$REMOVE` through `$let` in `$project`
- Cross-`$let` variable isolation
- `$let` across multiple documents with missing fields

Ref: #17


documentdb_tests/compatibility/tests/core/operator/expressions/misc/literal/test_literal_core.py
```diff
@@ -0,0 +1,527 @@
...
+        expected=Regex("^abc", "i"),
+        msg="Should return regex as-is",
+    ),
+    ExpressionTestCase(
+        "javascript",
+        expression={"$literal": Code("function(){}")},
+        expected=Code("function(){}"),
+        msg="Should return javascript as-is",
+    ),
+    ExpressionTestCase(
```
sathvikprasad97, line 128: Can we add System variable test cases

documentdb_tests/compatibility/tests/core/operator/expressions/misc/let/test_let_invalid.py
```diff
@@ -0,0 +1,197 @@
...
+        error_code=LET_MISSING_VARS_ERROR,
+        msg="Missing vars should fail",
+    ),
+    ExpressionTestCase(
+        "missing_in",
+        expression={"$let": {"vars": {"x": 1}}},
+        error_code=LET_MISSING_IN_ERROR,
+        msg="Missing in should fail",
+    ),
+    ExpressionTestCase(
```
sathvikprasad97, line 51: can we add a test case, {"$let": None}

documentdb_tests/compatibility/tests/core/operator/expressions/misc/literal/test_literal_project_disambiguation.py
```diff
@@ -0,0 +1,104 @@
...
+                {"$project": {"_id": 0, "result": {"$eq": ["$price", {"$literal": "$1"}]}}}
+            ],
+            "cursor": {},
+        },
+    )
+    assertSuccess(
+        result, [{"result": True}], msg="$literal '$1' should match field containing '$1'"
+    )
+
+
```
sathvikprasad97: can we add a test case, {"$literal": -1} in $project

documentdb_tests/compatibility/tests/core/operator/expressions/misc/let/test_let_core.py
```diff
@@ -0,0 +1,473 @@
...
+@pytest.mark.parametrize("test", CORE_TESTS, ids=lambda t: t.id)
+def test_let_core(collection, test):
+    """Test $let core behavior: type passthrough, scoping, and expression types."""
+    result = execute_expression(collection, test.expression)
+    assert_expression_result(result, expected=test.expected, msg=test.msg)
+
+
+# ---------------------------------------------------------------------------
+# Insert-based tests
+# ---------------------------------------------------------------------------
```
sathvikprasad97, line 402: can we add a test case for $let where vars references $$CURRENT directly

sathvikprasad97 (changes requested):

danielfrankcom:
@sathvikprasad97 addressed comments here and from internal review.

I added new integration test files to cover the concerns which foll outside of `$let` or `$literal` testing scope, which now sit at a higher level. It seemed worth investing in these tests now as the semantics around these operators are not particularly simple, and unlikely to be covered by other pipeline stage testing efforts.

I don't have a ton of context for these operator tests as I'm not the original author, but hopefully I've addressed your concerns.

danielfrankcom:
Bit of churn from signing, recent force pushes are mostly unchanged from before. Only new commit swaps parameterized tests to use `pytest_params`. Fixes for comments are visible in the isolated commits before my last comment.

sathvikprasad97 (approved):

documentdb_tests/framework/error_codes.py
```diff
@@ -291,3 +291,9 @@
 REPLACE_FIND_TYPE_ERROR = 10503901
 REPLACE_REPLACEMENT_TYPE_ERROR = 10503902
 REPLACE_INPUT_TYPE_ERROR = 10503904
+
+# Aliases for $let/$literal test compatibility
+INVALID_FIELD_PATH_DOLLAR_IN_VARIABLE_ERROR = UNRECOGNIZED_EXPRESSION_ERROR  # 168
```
eerxuan: How about we remove the no duplicate rule, so different names for same error can be in one place.

documentdb_tests/compatibility/tests/core/operator/expressions/test_expressions_combination_variables.py
```diff
@@ -0,0 +1,285 @@
...
+                "in": {
+                    "$let": {"vars": {"x": "$$y", "y": "$$x"}, "in": {"$subtract": ["$$x", "$$y"]}}
+                },
+            }
+        },
+        expected=1,
+        msg="Swap: x=2, y=1, 2-1=1",
+    ),
+    ExpressionTestCase(
+        "triple_nested_scoping_result_8",
```
eerxuan: triple_nested and triple_nested_scoping_result_8 test
  essentially the same thing

documentdb_tests/framework/error_codes.py
danielfrankcom: Huh, I didn't notice these were here, good catch. I don't think we need the different names personally, the more general names should be applicable I would think, but I'll confirm.

eerxuan (approved):

PR #104, opened 2026-04-15
Add $convert tests

danielfrankcom:
This change adds `$convert` tests.

Covers:

- Target type identification (string names, numeric codes, object form)
- `binData` conversions and `subtype` handling
- `format` and `byteOrder` parameters
- `onError` and `onNull` behavior
- Null/missing input and validation skipping
- Conversion errors, invalid arguments, and size limits
- Expression arguments and document field references

The `$convert` operator is the general form behind `$toString`, `$toInt`, `$toDouble`, etc. The various `$to*` tests are already parametrized over both the native operator and the equivalent `$convert` expression, so the test coverage is completely identical and there is no need to duplicate those test cases. The tests included here are unique to `$convert`, covering options that are not applicable to the `$to*` operators.

Note: there is a possible MongoDB bug where converting a `Date` with year outside 0-9999 to `string` produces error code 18537 (`DateToStringYearRange`) that is not caught by `onError`, unlike all other conversion errors. The `date_str_year_err_on_error_caught_*` tests document the correct behavior where `onError` handles this case, and are expected to fail against a server with this bug.


danielfrankcom:
The 2 remaining CI failures are the expected `date_str_year_err_on_error_caught_*` failures.

Once #79 is merged I can mark these as expected failures for MongoDB, but I don't want to duplicate the framework effort for adding that machinery.

manojnks (approved):

eerxuan (approved):

PR #111, opened 2026-04-16
Added tests for $ifNull

vic-tsang:
This PR contains 
- tests for $ifNull 

Ref: Issue #17


documentdb_tests/compatibility/tests/core/operator/expressions/conditional/ifNull/test_ifNull_system_variables.py
```diff
@@ -0,0 +1,50 @@
+"""
+Tests for $ifNull with system variables.
```
fanyangv: The docstring says "Tests for $ifNull with system variables" and the file is named test_ifNull_system_variables.py, but the let_variable_non_null and let_variable_null test cases (lines 28–39) use $let with $$x, which is a user-defined variable, not a system variable. System variables are built-ins like $$CURRENT, $$ROOT. Consider renaming the file to test_ifNull_variables.py and updating the docstring to "Tests for $ifNull with system and user-defined variables" to accurately reflect the scope.

documentdb_tests/compatibility/tests/core/operator/expressions/conditional/ifNull/test_ifNull_evaluation_order.py
```diff
@@ -0,0 +1,69 @@
...
+from documentdb_tests.compatibility.tests.core.operator.expressions.utils.expression_test_case import (  # noqa: E501
+    ExpressionTestCase,
+)
+from documentdb_tests.compatibility.tests.core.operator.expressions.utils.utils import (  # noqa: E501
+    assert_expression_result,
+    execute_expression_with_insert,
+)
+from documentdb_tests.framework.error_codes import BAD_VALUE_ERROR
+
+SHORT_CIRCUIT_TESTS: list[ExpressionTestCase] = [
```
fanyangv: The group name SHORT_CIRCUIT_TESTS and the test ID short_circuit_divide_by_zero (line 20) are misleading. The first test expects error_code=BAD_VALUE_ERROR even though $a=1 is non-null — this proves $ifNull does not short-circuit; it eagerly evaluates all inputs. The msg on line 24 correctly says "evaluates all inputs", but the group/ID names imply the opposite. Consider renaming to something like EAGER_EVALUATION_TESTS / eager_eval_divide_by_zero so the names match the actual behavior being demonstrated.

documentdb_tests/compatibility/tests/core/operator/expressions/conditional/ifNull/test_ifNull_expressions.py
```diff
@@ -0,0 +1,98 @@
...
+
+DOTTED_PATH_TESTS: list[ExpressionTestCase] = [
+    ExpressionTestCase(
+        "dotted_path_resolves_to_array",
+        expression={"$ifNull": ["$a.b.c", "default"]},
+        doc={"a": [{"b": {"c": "1"}}, {"b": {"c": "2"}}]},
+        expected=["1", "2"],
+        msg="$ifNull with dotted path into array of objects should resolve to array of values",
+    ),
+    ExpressionTestCase(
```
fanyangv, line 77: The msg for dotted_path_array_with_missing_subfield seems like an unfinished sentence: "Dotted path into array with missing subfield returns". It should explain why the expected value is [] and not "default" — when a dotted path traverses into an array and the terminal field (d) is missing from all subdocuments, MongoDB returns [], and since [] is not null/missing, $ifNull returns it as-is. Suggested: "Dotted path into array with missing subfield returns empty array (not null), so $ifNull returns [] rather than default".

fanyangv (changes requested):

documentdb_tests/compatibility/tests/core/operator/expressions/conditional/ifNull/test_ifNull_system_variables.py
vic-tsang: renamed file and fixed doc string

documentdb_tests/compatibility/tests/core/operator/expressions/conditional/ifNull/test_ifNull_evaluation_order.py
vic-tsang: addressed inaccurate documentation

documentdb_tests/compatibility/tests/core/operator/expressions/conditional/ifNull/test_ifNull_expressions.py
vic-tsang: fixed

fanyangv (approved):

eerxuan (approved):

PR #114, opened 2026-04-16
Replace lambda id extractors with pytest_params in parametrized tests

danielfrankcom:
This change replaces all `ids=lambda t: t.id` patterns in `@pytest.mark.parametrize` decorators with calls to `pytest_params()`, which also passes through per-case marks.

Added a check to `test_format_validator` that rejects `ids=lambda` in parametrize decorators, which should help prevent this going forward.


documentdb_tests/framework/test_format_validator.py
```diff
@@ -17,7 +17,7 @@ def validate_test_format(file_path: str) -> list[str]:
     errors: list[str] = []
 
     try:
-        with open(file_path, "r") as f:
+        with open(file_path) as f:
```
danielfrankcom, line 20: Drive-by change here, `"r"` is the default mode already, flagged by the linter.

vic-tsang (approved):

eerxuan (approved):

PR #120, opened 2026-04-16
Add $rand expression tests

danielfrankcom:
This change adds tests for the `$rand` expression operator. These tests were originally written by [@mitchell-elholm](https://github.com/mitchell-elholm).

**Test files added:**
- Argument handling tests for empty array/object syntax, invalid argument types (all BSON types, field references, `$missing`), and non-empty array/object error codes
- Return value tests for type (double), range [0, 1), and per-invocation independence
- Combination tests covering arithmetic operators (`$multiply`, `$add`, `$subtract`, `$pow`, `$mod`, `$divide`), conditional operators (`$cond`), type conversions (`$toString`, `$concat`), `$let` variable binding and reuse, `$switch`, `$filter`, object expressions, and deep nesting
- Pipeline stage integration tests for `$project`, `$group`, and `$match` covering per-document independence, statistical distribution, uniform distribution, range validation, null field handling, and empty collection behavior — placed at operator level to match repo conventions

Ref: #17


documentdb_tests/compatibility/tests/core/operator/expressions/misc/rand/test_rand_return_value.py
```diff
@@ -0,0 +1,34 @@
...
+    execute_expression,
+)
+
+
+def test_rand_basic(collection):
+    """Test {$rand: {}} is >= 0.0 and < 1.0."""
+    result = execute_expression(collection, {"$rand": {}})
+    assert_expression_result(
+        result, expected=pytest.approx(0.5, abs=0.5), msg="Should return value in [0, 1)"
+    )
```
manojnks: │ `pytest.approx(0.5, abs=0.5)` accepts values in `[0.0, 1.0]` (inclusive upper bound), but `$rand` specifies `[0, 1)`. A value of exactly `1.0` would pass this assertion but would be a bug. The argument handling file already does this correctly with server-side `$gte`/`$lt` —  
  consider using the same approach here for consistency:                                                                                                                                                                                                                                 
  │                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                         
  > result = execute_expression(                                                                                                                                                                                                                                                         
  >     collection,                                                                                                                                                                                                                                                                      
  >     {                                                                                                                                                                                                                                                                                
  >         "$let": {                                                                                                                                                                                                                                                                    
  >             "vars": {"r": {"$rand": {}}},                                                                                                                                                                                                                                            
  >             "in": {"$and": [{"$gte": ["$$r", 0.0]}, {"$lt": ["$$r", 1.0]}]},                                                                                                                                                                                                         
  >         }                                                                                                                                                                                                                                                                            
  >     },                                                                                                                                                                                                                                                                               
  > )                                                                                                                                                                                                                                                                                    
  > assert_expression_result(result, expected=True)                                                                                                                                                                                                                                      
  >

documentdb_tests/compatibility/tests/core/operator/expressions/test_expressions_combination_rand.py
```diff
@@ -0,0 +1,288 @@
...
+            "$let": {
+                "vars": {"r": {"$mod": [{"$rand": {}}, 1]}},
+                "in": {"$and": [{"$gte": ["$$r", DOUBLE_ZERO]}, {"$lt": ["$$r", 1.0]}]},
+            }
+        },
+    )
+    assert_expression_result(result, expected=True, msg="Should produce value in [0, 1)")
+
+
+def test_rand_divide_one_by_rand(collection):
```
manojnks: If `$rand` returns exactly `0.0`, this is a division by zero.

danielfrankcom: Not sure if you saw it, but there is a `test_rand_two_calls_differ` test which has the same likelihood of causing a collision between 2 values as this does of generating `0.0`.

I think it's probably more valuable to keep `test_rand_two_calls_differ` despite this since it tests something important. For the divide test here we can just write it differently, there's no need to have it do exactly this operation.

manojnks (approved):

eerxuan (approved):

PR #126, opened 2026-04-20
Add $project stage tests

danielfrankcom:
This change introduces tests for the `$project` aggregation stage.

I also renamed three error code constants to better reflect their semantics. These weren't used elsewhere as they were added recently based on my first cut at these tests.


documentdb_tests/compatibility/tests/core/operator/stages/project/test_project_acceptance.py
```diff
@@ -0,0 +1,233 @@
...
+    StageTestCase,
+)
+from documentdb_tests.framework.assertions import assertResult
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+
+# Property [$meta in Projection]: $meta expressions are accepted in
+# projection, with some meta types producing no visible field and $meta
+# being the only expression form allowed in exclusion mode.
+PROJECT_META_TESTS: list[StageTestCase] = [
```
sathvikprasad97, line 20: can we add a test for $meta: "textScore" which is commonly used with text search

documentdb_tests/compatibility/tests/core/operator/stages/project/test_project_bson_types.py
```diff
@@ -0,0 +1,435 @@
...
+        msg="$project field path should preserve Binary",
+    ),
+    StageTestCase(
+        "field_path_regex",
+        docs=[{"_id": 1, "v": Regex("abc", "i")}],
+        pipeline=[{"$project": {"r": "$v"}}],
+        expected=[{"_id": 1, "r": Regex("abc", "i")}],
+        msg="$project field path should preserve Regex",
+    ),
+    StageTestCase(
```
sathvikprasad97, line 382: Test for $-prefixed nested keys in sub-projections

sathvikprasad97 (changes requested):

documentdb_tests/compatibility/tests/core/operator/stages/project/test_project_bson_types.py
danielfrankcom: I believe this is already covered by `expr_unrecognized_operator`

sathvikprasad97 (approved):

eerxuan (changes requested):
Cross checking with PR132, no test for not exist collection. collection.drop() then run command. This case is stages specific, so need in every stage. stage like $collStats will fail.

danielfrankcom:
> Cross checking with PR132, no test for not exist collection. collection.drop() then run command. This case is stages specific, so need in every stage. stage like $collStats will fail.

Nice catch, this a problem for all aggregation stage tests.

I've fixed this one with the recent push, but will make similar fixes to #130 and #131. There have been several stages merged without these tests, but I should be able to fix those with a combined PR.

eerxuan (approved):

PR #129, opened 2026-04-22
Add collection $drop tests

danielfrankcom:
This change adds tests for the `$drop` operator.

It introduces a new `CommandTestCase` class which is intended to be shared by all collection command tests. The big difference from existing classes like `ExpressionTestCase` is that collection command tests tend to need fixture data as part of the test case. To resolve this I've provided the ability to use a `Callable` to get information about the fixtures while defining the test case. Similar to `StageTestCase`, a `setup` hook is also provided to configure prerequisites like specifically named collections.

I also had to modify `assertResult` to allow `raw_res` to be punched through to `assertSuccess`.


documentdb_tests/compatibility/tests/core/collections/commands/drop/test_drop_response_fields.py
```diff
@@ -0,0 +1,132 @@
...
+            db["test_text"].create_index([("content", "text")]),
+            db["test_text"],
+        )[-1],
+        command={"drop": "test_text"},
+        expected=lambda ctx: {"nIndexesWas": 2, "ns": ctx.namespace, "ok": 1.0},
+        msg="Should succeed with text index",
+    ),
+    CommandTestCase(
+        "ttl",
+        setup=lambda db: (
```
sathvikprasad97: All 6 test functions have an identical 9-line body (setup → insert → ctx → execute → assert). can we Extract to a shared run_command_test(database_client, collection, test) helper in commands/utils/ to reduce duplication.

sathvikprasad97 (changes requested):

documentdb_tests/compatibility/tests/core/collections/commands/drop/test_drop_response_fields.py
danielfrankcom: In the latest revision, the scope of each test method should be reduced. Some of the shared logic has been moved to the dataclass to help share some of the logic. Communicated with @eerxuan offline.

danielfrankcom:
After offline discussion, we landed on an improved design for `CommandTestCase` which has been pushed here.

I changed `CommandTestCase` to replace the freeform `setup` lambda with a declarative `target_collection` field using typed `TargetCollection` subclasses (`ViewCollection`, `CappedCollection`, `NamedCollection`) for collection setup, and an `indexes` field accepting `pymongo.IndexModel` instances. The `TargetCollection` subclasses and index approach should be reusable across future command tests.

This eliminates the tuple-chaining setup lambdas, and also fixes parallel safety by deriving all collection names from the fixture rather than hardcoding them. The setup logic is now consolidated into a single `prepare()` method on the dataclass. The `command` and `expected` fields remain explicit so each test case is still self-describing.

Cases that don't fit the dataclass pattern (dropping the underlying collection of a view, dropping `system.views`) have been moved to standalone test functions.

sathvikprasad97 (approved):

eerxuan (approved):

PR #131, opened 2026-04-22
Add $count aggregation stage tests

danielfrankcom:
This change adds aggregation stage tests for `$count`.


documentdb_tests/compatibility/tests/core/operator/stages/count/test_count_errors.py
```diff
@@ -0,0 +1,440 @@
...
+        "arity_extra_key",
+        docs=[{"_id": 1}],
+        pipeline=[{"$count": "total", "extra": 1}],
+        error_code=PIPELINE_STAGE_EXTRA_FIELD_ERROR,
+        msg="$count should reject a stage document with multiple keys",
+    ),
+]
+
+# Property [Type Strictness]: any non-string type is rejected as the count field.
+COUNT_TYPE_STRICTNESS_TESTS: list[StageTestCase] = [
```
sathvikprasad97, line 50: can we also include UUID

documentdb_tests/compatibility/tests/core/operator/stages/count/test_count_errors.py
```diff
@@ -0,0 +1,440 @@
...
+    ),
+    StageTestCase(
+        "type_expression_object",
+        docs=[{"_id": 1}],
+        pipeline=[{"$count": {"$literal": "x"}}],
+        error_code=COUNT_FIELD_TYPE_ERROR,
+        msg="$count should reject expression-like object without evaluating it",
+    ),
+]
+
```
sathvikprasad97, line 178: Can we add a test case, special-float rejection cases: NaN, +/-Infinity, -0.0.

sathvikprasad97 (changes requested):

documentdb_tests/compatibility/tests/core/operator/stages/count/test_count_errors.py
danielfrankcom: I'm not sure we need this, since it's not a native type, and is just another variant of binary which is already covered by the `type_binary` case.

We don't include UUIDs for any other BSON type enumerations in existing tests, and only use it when the operator has specific handling for binary, which this one doesn't.

danielfrankcom: I just did some more poking around and I see that both the `$limit` and `$sample` tests have this UUID variant which I assume is where this request came from.

Given our other tests before these newly introduced ones don't do this, my preference would be to remove them from both of those tests too given the reasons I stated above.

I can add them if you feel strongly about it, but note this is not something we have been doing in any existing tests that enumerate BSON types other than those 2 new ones (which I think we should remove).

documentdb_tests/compatibility/tests/core/operator/stages/count/test_count_errors.py
danielfrankcom: Similar to [here](https://github.com/documentdb/functional-tests/pull/131#discussion_r3134732137), we have not been enumerating these special numeric types for scenarios where the operator does not expect a numeric.

It's useful to cover these when any kind of numeric value is accepted and processed, but not so much when the operator only accepts a string like here, and rejects everything else. Given we already have testing for floats in this list, I don't think this is necessary.

documentdb_tests/compatibility/tests/core/operator/stages/count/test_count_errors.py
danielfrankcom: I updated #137 and #138 to remove these for consistency with past tests.

sathvikprasad97 (approved):

eerxuan (approved):

PR #138, opened 2026-04-24
Add $sample aggregation stage tests

danielfrankcom:
This change adds tests for the `$sample` aggregation stage operator.

Due to the random nature of `$sample` this is a little harder to test than other aggregation stages. I've worked around this by splitting testing into the following 2 approaches where appropriate:
- Using `$count` to focus on the number of documents, without caring which were chosen
- Using `"size": 1` to focus on the structure of the output, without caring how many were chosen

I have also introduced a `test_sample_statistical.py` file for this operator which has some basic checks for the content/distribution of the sampling. Being random, these tests could technically fail, but I've tried to structure them such that the probability of a false negative is so low that for all intents and purposes it's impossible.


documentdb_tests/compatibility/tests/core/operator/stages/sample/test_sample_statistical.py
```diff
@@ -0,0 +1,83 @@
...
+
+import pytest
+
+from documentdb_tests.framework.executor import execute_command
+
+
+# Property [Randomness]: repeated executions of the same $sample pipeline may
+# return different documents.
+@pytest.mark.aggregate
+def test_sample_statistical_randomness(collection):
```
sathvikprasad97, line 18: can you add a brief comment explaining why we don’t use the StageTestCase/assertResult pattern?

sathvikprasad97 (changes requested):

sathvikprasad97 (approved):

eerxuan (approved):

PR #144, opened 2026-04-27
Add $collStats aggregation stage tests

danielfrankcom:
This change adds tests for the `$collStats` operator.

Due to the nature of the operator, I needed to take a different approach to the assertions here. We can't reliably capture the entire expected shape of the result, so I added some machinery for asserting properties of the result. This should let us provide the coverage we need for `$collStats`, but also be applicable to future operators with complex results structures.

I also tried to avoid implementation-specific logic, since these stats seem to expose different things depending on the engine internals. The tests here are focused on the operator spec as best as possible.

I had to move `strict_equal` to its own module to avoid a circular import issue. The content is exactly the same, I just pulled it out.


documentdb_tests/compatibility/tests/core/operator/stages/collStats/test_collStats_type_errors.py
```diff
@@ -0,0 +1,138 @@
...
+    (Code("function(){}"), "code"),
+    (Code("function(){}", {}), "code_with_scope"),
+    (MinKey(), "minkey"),
+    (MaxKey(), "maxkey"),
+]
+
+# Property [histograms Sub-Field Type Strictness]: latencyStats.histograms
+# accepts only boolean true or false - all non-boolean types produce
+# TYPE_MISMATCH_ERROR, including null.
+HISTOGRAMS_TYPE_ERROR_TESTS: list[CollStatsTestCase] = [
```
sathvikprasad97, line 49: This excludes only True but should also exclude False; False is a valid boolean.

documentdb_tests/compatibility/tests/core/operator/stages/collStats/test_collStats_scale.py
```diff
@@ -0,0 +1,426 @@
...
+        coll, {"aggregate": coll.name, "pipeline": test.pipeline, "cursor": {}}
+    )
+    test.assert_result(result)
+
+
+# Property [Scale Factor Divides Size Fields]: specifying storageStats.scale=N
+# divides size fields (size, storageSize, totalSize, totalIndexSize, each entry
+# in indexSizes) by N using floor division, while avgObjSize, count, nindexes,
+# capped, and scaleFactor are unaffected.
+SCALE_AFFECTED_FIELDS = [
```
sathvikprasad97, line 336: "indexSizes" is missing here

documentdb_tests/compatibility/tests/core/operator/stages/collStats/test_collStats_validation.py
```diff
@@ -0,0 +1,288 @@
...
+            "queryExecStats": Exists(),
+        },
+        msg="sub-option ordering should not affect output",
+    ),
+]
+
+# Property [Multiple Options - Duplicate Keys]: when duplicate keys are present
+# in the argument document, the option is enabled if any occurrence has a
+# non-null value, regardless of key ordering.
+DUPLICATE_KEY_TESTS: list[CollStatsTestCase] = [
```
sathvikprasad97, line 100: can we include few more tests other then count

documentdb_tests/compatibility/tests/core/operator/stages/collStats/test_collStats_type_errors.py
```diff
@@ -0,0 +1,138 @@
...
+
+# Property [Stage Document Extra Keys]: a pipeline stage document with extra
+# keys alongside $collStats produces PIPELINE_STAGE_EXTRA_FIELD_ERROR.
+STAGE_EXTRA_KEY_TESTS: list[CollStatsTestCase] = [
+    CollStatsTestCase(
+        "extra_key",
+        docs=[{"_id": 1}],
+        pipeline=[{"$collStats": {}, "extra": 1}],
+        error_code=PIPELINE_STAGE_EXTRA_FIELD_ERROR,
+        msg="extra key in stage document should produce error 40323",
```
sathvikprasad97: Is this the correct error message?

sathvikprasad97 (changes requested):

documentdb_tests/compatibility/tests/core/operator/stages/collStats/test_collStats_type_errors.py
danielfrankcom: This test should just be removed actually, it's not really in the scope of testing `$collStats` since it's a sibling behavior.

Also noticed a few error codes snuck through in error messages, have removed those in the upcoming push.

documentdb_tests/compatibility/tests/core/operator/stages/collStats/test_collStats_scale.py
danielfrankcom: This is handled explicitly in the test method since it's not a scalar and can't be treated the same way:

```python
    for key, val in b["indexSizes"].items():
        checks[f"storageStats.indexSizes.{key}"] = Eq(int(val // scale))
```

sathvikprasad97 (approved):

eerxuan (approved):
The check machinery is neat. One follow up is we should only test DocumentDB spec. Especially storage related fields. But since DocumentDB spec is not defined yet. Creating issue to discuss https://github.com/documentdb/functional-tests/issues/160

eerxuan (approved):

PR #146, opened 2026-04-28
Add $bucket aggregation stage tests

danielfrankcom:
This change defines tests for the `$bucket` operator, following the existing pattern defined by other aggregation stage operators.


documentdb_tests/compatibility/tests/core/operator/stages/bucket/test_bucket_boundaries.py
```diff
@@ -0,0 +1,292 @@
...
+            {"_id": pytest.approx(math.nan, nan_ok=True), "count": 1},
+            {"_id": 0, "count": 1},
+        ],
+        msg="$bucket should accept float NaN as lowest boundary",
+        id="float_nan_boundary",
+    ),
+    StageTestCase(
+        docs=[{"_id": 1, "x": FLOAT_NEGATIVE_NAN}, {"_id": 2, "x": 5}],
+        pipeline=[{"$bucket": {"groupBy": "$x", "boundaries": [FLOAT_NEGATIVE_NAN, 0, 10]}}],
+        expected=[
```
sathvikprasad97, line 127: Will the assertion fail here?

documentdb_tests/compatibility/tests/core/operator/stages/bucket/test_bucket_accumulator_operators.py
```diff
@@ -0,0 +1,473 @@
...
+        pipeline=[
+            {
+                "$bucket": {
+                    "groupBy": "$x",
+                    "boundaries": [0, 10],
+                    "output": {"r": {"$stdDevPop": "$v"}},
+                }
+            }
+        ],
+        expected=[{"_id": 0, "r": 8.16496580927726}],
```
sathvikprasad97, line 138: Can the framework handle approximations of floating-point numbers in different engines?

documentdb_tests/compatibility/tests/core/operator/stages/bucket/test_bucket_boundaries.py
danielfrankcom: No, this is correct. The `pytest.approx` method compares all float NaN as equivalent

documentdb_tests/compatibility/tests/core/operator/stages/bucket/test_bucket_accumulator_operators.py
danielfrankcom: This is the IEEE 754 double-precision result for this operation, which should be consistent across implementations. If an engine handles this differently then it's a compatibility difference we would want to catch by this test failing.

sathvikprasad97 (approved):

eerxuan (approved):

PR #148, opened 2026-04-29
Add $lookup aggregation stage tests

danielfrankcom:
This change adds tests for the `$lookup` aggregation stage operator.

As part of this change I had to add a new `LookupTestCase` class which extends `StageTestCase`, in order to deal with some complexity that is specific to this operator. The tests need a way to define and reference foreign documents, which means we need to create those documents as part of the test run, and replace a foreign reference sentinel is the actual pipeline definition.


documentdb_tests/compatibility/tests/core/operator/stages/utils/stage_test_case.py
```diff
@@ -18,7 +18,7 @@ class StageTestCase(BaseTestCase):
     """Test case for pipeline stage tests."""
 
     docs: list[dict[str, Any]] | None = None
-    pipeline: list[dict[str, Any]] | None = None
+    pipeline: list[dict[str, Any]] = field(default_factory=list)
```
sathvikprasad97, line 21: Are there any existing StageTestCase subclasses that verify that the pipeline is None?

documentdb_tests/compatibility/tests/core/operator/stages/lookup/test_lookup_error_precedence.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@pytest.mark.aggregate
+@pytest.mark.parametrize("test_case", pytest_params(LOOKUP_ERROR_PRECEDENCE_TESTS))
+def test_lookup_error_precedence(
+    collection: Any,
+    test_case: LookupPrecedenceTestCase,
+) -> None:
+    """Test $lookup error precedence ordering."""
+    populate_collection(collection, test_case)
```
sathvikprasad97: Should we use the setup_lookup function ?

documentdb_tests/compatibility/tests/core/operator/stages/utils/stage_test_case.py
danielfrankcom: No, it's not a valid configuration to provide `None`. I think it was provided originally because every field here has to have a default (result of being stuck with Python 3.9 compatibility atm), but `None` isn't a sensible choice here.

This change just helps to avoid unnecessary checking for `None` in consumers of this field, since we can treat it as always being a list after this change.

documentdb_tests/compatibility/tests/core/operator/stages/lookup/test_lookup_error_precedence.py
danielfrankcom: Found a different issue while investigating this which invalidates the need for the custom class here, so this now looks the same as the other ones.

sathvikprasad97 (approved):

documentdb_tests/compatibility/tests/core/operator/stages/lookup/test_lookup_error_precedence.py
```diff
@@ -0,0 +1,130 @@
+"""Tests for $lookup error precedence ordering."""
```
eerxuan: We don't want to test these. It can be different on each engine. We need to add this rule in the agent prompts.

documentdb_tests/compatibility/tests/core/operator/stages/lookup/test_lookup_let_errors.py
```diff
@@ -0,0 +1,592 @@
...
+                }
+            }
+        ],
+        error_code=LET_UNDEFINED_VARIABLE_ERROR,
+        msg="$lookup should reject cross-referencing between sibling let variable definitions",
+    ),
+]
+
+# Property [let Cluster Time Error]: $$CLUSTER_TIME as a let value
+# produces an error because it is not available in standalone mode.
```
eerxuan: This is not testing $lookup

eerxuan (changes requested):

documentdb_tests/compatibility/tests/core/operator/stages/lookup/test_lookup_let_errors.py
danielfrankcom: Agreed, will remove

documentdb_tests/compatibility/tests/core/operator/stages/lookup/test_lookup_result_size.py
```diff
@@ -0,0 +1,72 @@
...
+                    "from": FOREIGN,
+                    "localField": "lf",
+                    "foreignField": "ff",
+                    "as": "joined",
+                }
+            },
+            {"$project": {"joined_count": {"$size": "$joined"}}},
+        ],
+        expected=[{"_id": 1, "joined_count": 20}],
+        msg="$lookup should produce a joined array that exceeds 16MB",
```
ramgokul-14, line 55: Not sure whether to put this test here or somewhere else but another test - for same data, if lookup + unwind is used it should not error out

ramgokul-14 (commented):
I think it would not be part of this specific PR, but $lookup + $unwind together needs extensive tests. Just a call out

danielfrankcom:
> I think it would not be part of this specific PR, but $lookup + $unwind together needs extensive tests. Just a call out

Agreed, I'm not sure deep testing of this belongs here, but I did add a representative test of the 16MB limit to the integration test file.

eerxuan (approved):

PR #152, opened 2026-04-30
Shared test constants and helpers for argument handling coverage

eerxuan:
Replaces #54. Extracts argument count and type validation into shared constants and helpers.

**Framework changes:**
- `ErrorCode(int)` subclass in `error_codes.py` — distinct type for error codes so assertions can auto-route
- `assertResultV2` in `assertions.py` — single `expected` param, routes `ErrorCode` to failure assertion automatically. `assertResult` now delegates to it.
- `assert_expression_result_v2` in `utils.py` — expression wrapper that auto-wraps success values in `[{"result": ...}]`
- `with_expected()` in `parametrize.py` — pairs shared input samples with per-operator expected outcomes

**Shared test constants in `test_constants.py`:**
- `BSON_TYPE_SAMPLES` — all 19 BSON types for exhaustive type validation
- `ARRAY_INPUT_ARGS` — arg counts 0–5 + representative non-array types for arity testing
- `OBJECT_INPUT_ARGS` — representative non-object types + invalid object shapes (`{}`, `{invalid_field: None}`)

**Demo:**
- `test_operator_divide_argument_handling.py` — 49 tests using the shared pattern (5 arg count + 19 dividend type + 19 divisor type + 6 non-array type)
- Removed replaced inline type tests from `test_operator_divide.py`


documentdb_tests/framework/test_constants.py
```diff
@@ -189,3 +190,54 @@
...
+    pytest.param(MaxKey(), id="maxkey"),
+    pytest.param(b"\x00", id="bindata"),
+    pytest.param(None, id="null"),
+]
+
+# Argument lists of length 0–5 and representative non-array types
+# for testing array-input operators.
+# Uses field refs to missing fields (type-neutral) so arg-count validation
+# triggers before any type check, regardless of operator input type.
+ARRAY_INPUT_ARGS = [
```
danielfrankcom, line 222: The non-array entries here (`"$a"`, `True`, `None`, `{}`, `{"x": 1}`) all produce the same arity error as `1_args`. The server treats any non-array value as a single argument:

```
string: code=16020 msg=Expression $divide takes exactly 2 arguments. 1 were passed in.
bool:   code=16020 msg=Expression $divide takes exactly 2 arguments. 1 were passed in.
```

So they're testing the same property as `1_args` (wrong argument count) rather than type rejection. If the intent is to cover non-array types being passed where an array is expected, that overlaps with what `BSON_TYPE_SAMPLES` already does.

I think the arity cases and the type-coverage cases are better kept separate since they're testing different validation layers.

documentdb_tests/framework/assertions.py
```diff
@@ -228,10 +229,45 @@ def assertFailure(
 
 def assertFailureCode(result: Union[Any, Exception], expected_code: int, msg: Optional[str] = None):
     """Assert command failed and check only the code field."""
-    expected = {"code": expected_code}
+    expected = {"code": int(expected_code)}
     assertFailure(result, expected, msg, transform=partial_match(expected))
 
 
+def assertResultV2(
```
danielfrankcom, line 236: Rather than adding `assertResultV2` and `assert_expression_result_v2` as new functions, could the existing `assertResult` and `assert_expression_result` handle this directly? Since `ErrorCode` is a subclass of `int`, you could add an `isinstance` check at the top of each existing function to route it:

```python
def assertResult(result, expected=None, error_code=None, ...):
    if isinstance(expected, ErrorCode):
        error_code = expected
        expected = None
    # ... rest unchanged
```

That keeps backward compatibility for existing `error_code=` call sites, supports passing `ErrorCode` as `expected` for the new pattern, and avoids maintaining parallel V2 functions.

Also a small thing, `INVALID_DOLLAR_FIELD_PATH` in `error_codes.py` wasn't wrapped in `ErrorCode(...)`.

documentdb_tests/framework/test_constants.py
```diff
@@ -189,3 +190,54 @@
...
+    pytest.param(True, id="bool"),
+    pytest.param(None, id="null"),
+    pytest.param({}, id="object"),
+    pytest.param({"x": 1}, id="non_empty_object"),
+]
+
+# Invalid arguments for object-input operators, stages, or commands
+# (e.g. $filter, $match, $createCollection).
+# Tests representative non-object types and invalid object shapes.
+OBJECT_INPUT_ARGS = [
```
danielfrankcom, line 235: This mixes two different properties, similar to `ARRAY_INPUT_ARGS`. The type-rejection portion overlaps with what `BSON_TYPE_SAMPLES` is meant to solve exhaustively. I think separating these into distinct groups would be clearer, one for non-object type rejection and one for invalid object shapes.

danielfrankcom (changes requested):
These could be separate comments but they're all related to the same design concerns so I figured it's easier to discuss them together.

I like the goal of having a canonical BSON type list that ensures complete coverage and reduces boilerplate, that seems like a problem worth solving to me. That being said, I have some concerns about the readability tradeoffs at call sites with the current design.

### Separating input from expected obscures the test logic

In the `$divide` test, the expected dict has `"int32": 0.5`. That result comes from dividing the sample value (`1`) by the test's divisor (`2`), but you can't see that relationship without going to `test_constants.py` to look up what `"int32"` maps to. With this approach, the reader has to hold two files in their head to understand why `0.5` is correct. Compare to the existing style in the same file:

```python
DivideTest(
    "same_type_int32", dividend=10, divisor=2, expected=5.0, ...
),
```

When the expected outcome is a computation over the input, co-locating them makes the test self-documenting. The indirection through string IDs works against that.

### The string-keyed dict has no completeness enforcement

Nothing validates that the expected dict covers exactly the set of sample IDs. A typo like `"decimall128": ERR` becomes a dead entry that never runs. An extra key that doesn't correspond to any sample would also be silently ignored.

Additionally, a missing key shows up as a `KeyError` at runtime. This could be made clearer with an explicit check and helpful message.

### Accepted types don't benefit from the shared sample list

The dict includes entries for accepted types like `"int32": 0.5` alongside the rejected ones, but a single representative value from a shared list doesn't provide meaningful coverage for an accepted type. If there's already dedicated testing for how `$divide` handles numeric inputs (which there is in this file), these BSON entries for accepted types are redundant. If there isn't dedicated testing, a single sample value from this BSON enumeration isn't sufficient to fill that gap.

Explicitly excluding the accepted types from the rejection test would make the coverage clearer imo. Either dedicated tests exist for those types, or the exclusion makes it obvious they're missing. Mixing them into the same dict with a computed expected value obscures both cases.

### The dict requires exhaustive specification even when most entries are identical

For the common case where most types produce the same error, you still have to write out 15+ entries that all say `ERR`. The fact that an `ERR` alias was introduced to make this tolerable is a pointer to this. It also obscures what's actually being asserted at each entry, since the reader has to look up what `ERR` maps to.

Additionally, though rare, if a new type is added to `BSON_TYPE_SAMPLES`, every call site using `with_expected()` needs a new entry added manually.

Filtering the list instead gives you both conciseness and automatic coverage of new types. For example, the dividend type-rejection test could be:

```python
DIVIDEND_TYPE_ERROR_TESTS = [
    DivideTest(
        f"dividend_type_{sample.id}",
        dividend=sample.value,
        divisor=2,
        error_code=TYPE_MISMATCH_ERROR,
        msg=f"$divide should reject {sample.id} as dividend",
    )
    for sample in bson_types_except("int32", "int64", "double", "decimal128", "null")
    # Could also provide constants like bson_types_except(BSON_NUMERIC, "null")
]
```

The specific value used for the dividend doesn't matter here since the operator rejects it before computing anything, so there are no magic values to trace back. It's also data-driven and doesn't need its own test method. You can tell what's being tested just from the definition.

PR #158, opened 2026-05-01
Update Test Coverage to exclude stages test from query tests

vic-tsang:
This coverage guideline update is to make sure we don't apply pipeline/stages tests to query predicate operators.


eerxuan (commented):
* I feel "Query predicate operators (e.g., `$mod`, `$regex`, `$in`, `$gt`) do not need `$expr` tests" is too specific. Just a reminder that query operators under operators/query/ are different. Which implies these changes.
* Also can you add these global rules at the top:
   1. No error precedence tests — Don't test which error fires first when multiple
   params are invalid. Each error should be tested individually.
   2. One layer deep only — Test the feature itself and one layer of composition.
   $project with expressions: one test per operator, no edge cases of the operator.
   $lookup sub-pipeline: verify stages are accepted/rejected, don't test the stage's full behavior inside the sub-pipeline.
   3. No engine-specific implementation details — Don't test exact nesting limits (20
   levels), pipeline stage limits (1000), or internal storage details (wiredTiger).
   These belong in engine-specific test suites, not compatibility tests.
   4. Tests belong in the feature's folder — $abs type validation goes in
   expressions/arithmetic/abs/, not in stages/project/. Testing $abs inside $project
   is a $project context test (one simple case), not an $abs test.

eerxuan (approved):

Issue #26, opened 2026-04-03
Add compatibility test for core: collections (16 total)


eerxuan:
core	collections	-	-	-	capped-collections
core	collections	-	-	-	clustered-collections
core	collections	-	-	-	materialized-views
core	collections	-	-	-	timeseries
core	collections	-	-	-	views
core	collections	commands	-	-	cloneCollectionAsCapped
core	collections	commands	-	-	collMod
core	collections	commands	-	-	compact
core	collections	commands	-	-	convertToCapped
core	collections	commands	-	-	create
core	collections	commands	-	-	drop
core	collections	commands	-	-	dropDatabase
core	collections	commands	-	-	listCollections
core	collections	commands	-	-	listDatabases
core	collections	commands	-	-	renameCollection
core	collections	commands	-	-	validateDBMetadata

PR #161, opened 2026-05-04
Add $listDatabases command tests

danielfrankcom:
This change adds tests for `$listDatabases`. The intent here is not to test the functionality of the operator, but just the spec, meaning the parameters/config it accepts, and the results it produces. Verifying the actual database listing through creating/deleting them is considered out of scope for these tests.

As part of this change I defined common expectation formats which are reused by a bunch of tests, as they expect pretty much the same thing in most cases. As many tests are focused on parameter structure and content, the expected result is identical in a lot of places. This helps to reduce some duplication across the files.

Ref: #26


documentdb_tests/compatibility/tests/core/collections/commands/listDatabases/test_listDatabases_command_acceptance.py
```diff
@@ -0,0 +1,409 @@
...
+# Property [Null and Missing Behavior]: null literal and omitted
+# parameters are accepted and produce a successful response.
+NULL_MISSING_TESTS: list[CommandTestCase] = [
+    CommandTestCase(
+        command={"listDatabases": None},
+        expected=basic_success,
+        msg="Null command field value should be accepted",
+        id="command_value_null",
+    ),
+    CommandTestCase(
```
SaiPradyumn, line 52: can we have a test case covering  `maxTimeMS: null`

documentdb_tests/compatibility/tests/core/collections/commands/listDatabases/test_listDatabases_filter_predicates.py
```diff
@@ -0,0 +1,373 @@
...
+            "filter": functools.reduce(
+                lambda inner, _: {"$and": [inner]}, range(99), dict[str, Any]({"name": "admin"})
+            ),
+        },
+        expected={"ok": Eq(1.0), "databases": Contains("name", "admin")},
+        msg="Nesting depth up to 99 levels should be accepted",
+        id="filter_nesting_99",
+    ),
+]
+
```
SaiPradyumn, line 327: Can we have test cases to test Filter with unsupported query operators, let's check that with the team if we want to have that included in the framework.

SaiPradyumn (changes requested):

documentdb_tests/compatibility/tests/core/collections/commands/listDatabases/test_listDatabases_nameonly.py
```diff
@@ -0,0 +1,411 @@
...
+    INT32_MIN,
+    INT64_MAX,
+    INT64_MIN,
+    INT64_ZERO,
+)
+
+# Property [nameOnly Response Structure]: when nameOnly is true, the
+# response contains only databases and ok, and each database entry
+# contains only the name field.
+name_only_success_TESTS: list[CommandTestCase] = [
```
SaiPradyumn: can we please change this to NAME_ONLY_SUCCESS_TEST

SaiPradyumn (approved):

eerxuan (commented):
### 🟡 [Suggestion] `filter_nesting_99` doesn't bracket the actual nesting limit

  ```python
  "filter": functools.reduce(
      lambda inner, _: {"$and": [inner]}, range(99), {"name": "admin"}
  ),

  This wraps {"name": "admin"} in 99 nested $and levels and asserts success. The msg says "up to 99 levels should be accepted." But it doesn't:
  - Establish what the actual filter nesting limit is (BSON limit is 180, but query parser may have its own)
  - Show rejection at the limit + 1

  Compare to PR #167's key_depth_180_accepted / key_depth_181 pair which brackets the boundary clearly. Suggest either:
  - Drop the test (a single accepted-deep-nesting case provides little signal)
  - Or bracket the limit: pick the actual filter nesting limit, test accepted at limit and rejected at limit+1

  🟡 [Suggestion] Trim exhaustive comment parameter matrix

  test_listDatabases_comment.py has ~38 cases covering subnormal doubles, BOM strings, ZWJ sequences, 15MB strings, deeply nested objects, MinKey/MaxKey, etc. for the comment parameter.

  comment is a simple shared parameter that exists across many commands (listDatabases, listCollections, find, aggregate, count, distinct, update, etc.) and accepts any BSON type per the spec. A handful of representative cases (~5-8: string, document, number, array, null, ObjectId) is sufficient — once "any BSON type is accepted" is proven, additional encoding edge cases
  (BOM, ZWJ, 15MB strings, subnormals) don't add new signal.

eerxuan (approved):

eerxuan (approved):

PR #166, opened 2026-05-06
Add $listClusterCatalog stage tests

danielfrankcom:
This change adds tests for the `$listClusterCatalog` aggregation stage.

I modified `assertProperties` to run against all input documents. Previously it was only used by operators which produced 1 document, so this change is compatible with those existing cases.

I also introduced `ListClusterCatalogTestCase` for these cases, as there were some special requirements around collection handling and a few other things which aren't required by the existing `StageTestCase` users.

Ref: #34


documentdb_tests/compatibility/tests/core/operator/stages/listClusterCatalog/test_listClusterCatalog_output.py
```diff
@@ -0,0 +1,429 @@
...
+    ListClusterCatalogTestCase(
+        id="options_validator",
+        target_collection=CustomCollection(
+            options={"validator": {"$jsonSchema": {"bsonType": "object"}}},
+        ),
+        pipeline=lambda ctx: [{"$listClusterCatalog": {}}, {"$match": {"ns": ctx.ns}}],
+        expected={"options.validator": Exists()},
+        msg="Collection with validator should have validator in options",
+    ),
+    ListClusterCatalogTestCase(
```
SaiPradyumn, line 161: balancingConfiguration: true Parameter is validated for type errors (17 cases) but its actual behavior is never tested. Core gap.

documentdb_tests/compatibility/tests/core/operator/stages/listClusterCatalog/test_listClusterCatalog_output.py
```diff
@@ -0,0 +1,429 @@
...
+        id="options_validation_level_and_action",
+        target_collection=CustomCollection(
```
SaiPradyumn, line 163: balancingConfiguration: false / omitted test cases

documentdb_tests/compatibility/tests/core/operator/stages/listClusterCatalog/test_listClusterCatalog_output.py
```diff
@@ -0,0 +1,429 @@
...
+    CustomCollection,
+    ExistingDatabase,
+    NamedCollection,
+    TimeseriesCollection,
+    ViewCollection,
+)
+
+# Property [Default Output Fields]: every output document contains all
+# default fields.
+DEFAULT_FIELD_TESTS: list[ListClusterCatalogTestCase] = [
```
SaiPradyumn, line 28: sharded field value : (in DEFAULT_FIELD_TESTS — change "sharded":  Exists() to "sharded": Eq(False), or add a separate test

SaiPradyumn (approved):

eerxuan (approved):

PR #167, opened 2026-05-06
Add $geoNear stage tests

danielfrankcom:
This change adds tests for the `$geoNear` aggregation stage operator.

As part of this change I introduced an `indexes` property to `StageTestCase` since I have needed this for a few operators in this category now.

I also introduced a `foreign_indexes` property to `LookupTestCase`, and defined the integration test for `$lookup` and `$geoNear` on the existing `test_stages_position_lookup.py` file, since it has the necessary infrastructure to deal with foreign documents already. This avoided complicating the `$geoNear` integration test file.

Ref: #34


eerxuan (commented):
## Summary

Excellent, comprehensive `$geoNear` stage test suite covering 17 test files with thorough parametrized coverage of type errors, distance calculations, path validation, query operators, pipeline position constraints, and multi-stage combinations. The PR also introduces useful infrastructure changes (`indexes` on `StageTestCase`, `foreign_indexes` on `LookupTestCase`) that benefit future stage tests.

Overall this is a high-quality contribution that follows project conventions closely.

## Critical Issues

### 🔴 `no_collection` test case may have incorrect setup

In `test_geoNear_prerequisite_errors.py`, the test case `no_collection` uses `docs=None` with the expectation of `NAMESPACE_NOT_FOUND_ERROR`:

```python
StageTestCase(
    "no_collection",
    docs=None,
    pipeline=[...],
    error_code=NAMESPACE_NOT_FOUND_ERROR,
    ...
)
```

However, `populate_collection()` unconditionally calls `collection.database.create_collection(collection.name)` before checking `test_case.docs`. This means the collection **will exist** (just empty, with no indexes) when the aggregate runs, which should produce `INDEX_NOT_FOUND_ERROR` rather than `NAMESPACE_NOT_FOUND_ERROR`.

If this passes in CI, it would be worth understanding why — perhaps the fixture teardown/setup creates a different scenario. Otherwise, `populate_collection` needs a guard like:

```python
if test_case.docs is not None:
    collection.database.create_collection(collection.name)
    collection.insert_many(test_case.docs) if test_case.docs else None
```

Or the test should use a separate collection name that was never created.

## Suggestions

### 🟡 [Suggestion] Consider `FIELD_PATH_EMPTY_COMPONENT_ERROR` vs `FIELD_PATH_TRAILING_DOT_ERROR` distinction

In `test_geoNear_path_errors.py`, both "leading dot" and "consecutive dots" map to `FIELD_PATH_EMPTY_COMPONENT_ERROR`, while "trailing dot" maps to `FIELD_PATH_TRAILING_DOT_ERROR`. This is fine if that's what MongoDB does, just noting it for coverage awareness — the distinction between trailing dot being a separate error code vs other empty components is correctly captured.

### 🟢 [Nitpick] `_geonear_query` helper could have a docstring clarification

The helper in `test_geoNear_query_operators.py`:
```python
def _geonear_query(query: dict) -> list[dict]:
    """Build a $geoNear pipeline with the given query filter."""
```

This is well-placed (single file, one layer), keeping it local is the right call here.

### 🟢 [Nitpick] `reduce`/`cast` pattern for deep nesting

The pattern in `test_geoNear_output_fields.py`:
```python
**reduce(lambda v, k: {k: v}, reversed(["a"] * 199), cast(Any, DOUBLE_ZERO)),
```
is readable enough for the depth-199 boundary test — it's the most concise way to express this.

## Questions

1. For the `no_collection` test case — does the `collection` fixture skip `create_collection` or does `populate_collection` have additional branching not shown in this diff that handles `docs=None` as "don't create the collection"?

2. The `distanceField omitted` test expects documents with no `dist` field. On which MongoDB version was this validated? Older versions required `distanceField`; it became optional in 7.3+.

## Positive Feedback

- **Strict BSON types throughout**: `DOUBLE_ZERO` instead of `0.0`, `Int64`, `Decimal128` — all consistently applied, including the smoke test update.
- **Excellent property-based organization**: Every test list is prefixed with a `# Property [Name]: description` comment that maps directly to behavioral properties being tested.
- **Comprehensive error code coverage**: 12 new error codes added to `error_codes.py` with clean numerical ordering.
- **Smart infrastructure additions**: The `indexes` field on `StageTestCase` and `foreign_indexes` on `LookupTestCase` are minimal, backwards-compatible additions that unlock geo index testing without boilerplate.
- **Integration test placement is correct**: `test_stages_position_geoNear.py` at the stages level (not inside `geoNear/`) because it tests cross-stage composition, and the `$lookup`+`$geoNear` test reusing `test_stages_position_lookup.py` avoids duplicating foreign-collection infrastructure.
- **Query operator coverage is impressively thorough**: comparison, logical, element, evaluation, array, bitwise, and geospatial operators all validated within `$geoNear`'s query parameter.

eerxuan (commented):
## Summary

Comprehensive `$geoNear` stage test suite — 17 files covering type errors, distance calculations, path validation, query operators, pipeline position, and stage combinations. Infrastructure additions (`indexes` on `StageTestCase`, `foreign_indexes` on `LookupTestCase`) are clean and minimal.

I reviewed against TEST_FORMAT.md, TEST_COVERAGE.md, and FOLDER_STRUCTURE.md rule by rule. Format and coverage compliance is excellent. Folder structure has one notable issue around feature ownership.

## Critical Issues

None blocking.

## Suggestions

### 🟠 [Required] Field path validation tests violate feature ownership

Per **FOLDER_STRUCTURE.md rule 1**: *"Test only the feature's own inputs and edge cases, not lower-level features it contains (those are tested in their own directories)."*

`test_geoNear_path_errors.py` exhaustively tests generic field path validation — empty string, `$`-prefix, leading/trailing dot, consecutive dots, null byte, depth 200/201 — for each of `distanceField`, `includeLocs`, and `key`. These rules are enforced by MongoDB's general field path parser and apply equally to `$project`, `$set`, `$sort`, and many other operators. Existing tests in `stages/project/test_project_errors.py` (20 references), `stages/set/test_set_errors.py` (16), and `stages/sort/test_sort_spec_errors.py` (15) already cover these.

**What belongs in `$geoNear`** (keep these):
- `key_depth_181` → `GEO_NEAR_KEY_DEPTH_LIMIT_ERROR` (180-component limit) — `$geoNear`-specific stricter limit, genuinely valuable
- Non-constant expression rejections (`GEO_NEAR_NEAR_NOT_CONSTANT_ERROR`, `GEO_NEAR_MAX_DISTANCE_NOT_CONSTANT_ERROR`, `GEO_NEAR_MIN_DISTANCE_NOT_CONSTANT_ERROR`) — `$geoNear`-specific
- Dot-notation acceptance and numeric component literal-key semantics — `$geoNear` path *semantics*, not validation

**What should move out**: the empty/dollar-prefix/leading-dot/trailing-dot/consecutive-dots/null-byte/depth-200/depth-201 matrix per parameter. One smoke case per parameter ("`$geoNear` validates `distanceField`/`includeLocs`/`key` as field paths") would suffice here. The exhaustive matrix should live in a shared field-path test directory.

### 🟢 [Nitpick] `populate_collection` silently drops `indexes` when `docs=None`

In `stage_test_case.py`:

```python
def populate_collection(collection: Collection, test_case: StageTestCase) -> None:
    if test_case.docs is None:
        return  # ← indexes silently ignored if also set

    collection.database.create_collection(collection.name)
    if test_case.docs:
        collection.insert_many(test_case.docs)
    if test_case.indexes:
        collection.create_indexes(test_case.indexes)
```

If a future test writer sets `docs=None, indexes=[...]` (perhaps expecting an `INDEX_NOT_FOUND_ERROR` test scenario), the indexes silently disappear and the test passes for the wrong reason. No current test hits this combination, but it's a latent footgun.

Suggestion — fail loudly:
```python
if test_case.docs is None:
    if test_case.indexes:
        raise ValueError("indexes requires docs to not be None")
    return
```

Same applies to `LookupTestCase.foreign_indexes` if `foreign_docs=None`.

### 🟢 [Nitpick] `_geonear_query` helper is well-scoped

Single-file helper, single layer of abstraction over `execute_command` boilerplate — exactly what TEST_FORMAT says to do.

## Positive Feedback

- **Strict BSON types throughout**: `DOUBLE_ZERO`, `Int64`, `Decimal128`, `DECIMAL128_NAN`, `FLOAT_INFINITY` consistently used. Smoke test was correctly updated to use `DOUBLE_ZERO`.
- **Excellent property-based organization**: Every test list is prefixed with a `# Property [Name]: description` comment that maps directly to behavioral properties. This is the cleanest way to read what's covered.
- **Spherical coercion coverage is a standout**: tests truthy/falsy coercion across every BSON type using a generator pattern, covering BSON type distinction (rule 10) and demonstrating that omitted defaults to false. This is high-value coverage that's easy to miss.
- **Correct integration test placement**: `test_stages_position_geoNear.py` lives in the parent `stages/` folder per FOLDER_STRUCTURE rule 4. The `$lookup`+`$geoNear` test correctly goes into `test_stages_position_lookup.py` to reuse foreign-collection infrastructure.
- **Smart infrastructure additions**: `indexes` on `StageTestCase` and `foreign_indexes` on `LookupTestCase` are minimal, backwards-compatible, and unlock future geo tests.
- **`no_collection` test takes proper advantage of `populate_collection` semantics**: the documented `docs=None` → collection-not-created path tests namespace-not-found cleanly.
- **Stricter `key` depth limit (180 vs 200) is correctly tested as $geoNear-specific** — exactly the kind of test that belongs in `$geoNear/`.
- **Query operator coverage** (comparison, logical, element, evaluation, array, bitwise, geospatial) is thorough.
- **`$meta geoNearDistance`/`geoNearPoint` tests** are bonus coverage not currently mentioned in TEST_COVERAGE.md rule 13 — could prompt a docs update.

eerxuan (approved):

eerxuan (approved):

SaiPradyumn (approved):

nitinahuja89 (approved):

Issue #33, opened 2026-04-03
Add compatibility test for query: element and evaluation operators (7 total)


eerxuan:
core	operator	query	misc	-	exists
core	operator	query	misc	-	expr
core	operator	query	misc	-	jsonSchema
core	operator	query	misc	-	mod
core	operator	query	misc	-	regex
core	operator	query	misc	-	type
core	operator	query	misc	-	where

PR #169, opened 2026-05-07
Added query tests for $jsonSchema

vic-tsang:
This PR contains:
- query tests for $jsonSchema
- BSON type testing framework which tests out all the invalid BSON types 
  - the reason why we need to test valid BSON type as well for a parameter is to make sure the author doesn't mistakenly mark a parameter with a wrong valid type (in this case without the valid BSON type tests it's not easy to catch)
  - for valid BSON type, we are only checking if there is no error. We don't check if the operator return the correct value as other tests perform those tests

Ref:
- Issue #33


documentdb_tests/framework/test_constants.py
```diff
@@ -188,3 +189,52 @@
...
+
+
+# Representative sample value for each BSON type
+BSON_TYPE_SAMPLES = {
+    BsonType.DOUBLE: 3.14,
+    BsonType.STRING: "hello",
+    BsonType.OBJECT: {"key": "value"},
+    BsonType.ARRAY: ["a", "b", "c"],
+    BsonType.EMPTY_OBJECT: {},
+    BsonType.EMPTY_ARRAY: [],
```
danielfrankcom: Why do object and array need multiple variants in this list?

This seems like it is maybe encoding some testing of specific edge cases for the `$jsonSchema` operator which we should leave to dedicated tests. If we cover special values for objects/arrays, why not include infinity for float/Decimal128? It seems like a slippery slope.

documentdb_tests/compatibility/tests/core/operator/query/misc/jsonSchema/test_jsonSchema_keyword_validation.py
```diff
@@ -0,0 +1,282 @@
...
+from documentdb_tests.framework.bson_type_validator import (
+    BsonType,
+    BsonTypeTestCase,
+    generate_bson_acceptance_test_cases,
+    generate_bson_rejection_test_cases,
+)
+from documentdb_tests.framework.error_codes import BAD_VALUE_ERROR, FAILED_TO_PARSE_ERROR
+from documentdb_tests.framework.executor import execute_command
+
+JSONSCHEMA_PARAMS = [
```
danielfrankcom, line 21: I left a comment similar to this [here](https://github.com/documentdb/functional-tests/pull/152#pullrequestreview-4238040408), but I think we can potentially replace the complex case building machinery in `bson_type_validator.py`, and use a structure like the one below.

I specifically chose this subset to show how the special params like `error_codes_overrides`, `requires`, `default_error_code`, etc. fold into this structure. This should be applicable to all `jsonSchema` parameters we are trying to test.

The tradeoff is a few more lines of test data, but each list is self-documenting. I don't think this would be a dramatic amount more code, and it would be reusable/consistent with other operators in the code base. You can read the filter, error code, and message without jumping to `bson_type_validator.py` or understanding the `expected_code()` dispatch. Personally I think this may be a worthwhile tradeoff, though I recognize this is subjective.

 ```python
   # --- Rejection at root level ---

   REJECTION_ROOT_TESTS: list[QueryTestCase] = [
       QueryTestCase(
           id=f"reject_{bt.value}_for_required_root",
           filter={"$jsonSchema": {"required": sample}},
           error_code=FAILED_TO_PARSE_ERROR if bt == BsonType.EMPTY_ARRAY else TYPE_MISMATCH_ERROR,
           msg=f"required should reject {bt.value} at root",
       )
       for bt, sample in BSON_TYPE_SAMPLES.items()
       if bt not in {BsonType.ARRAY}
   ] + [
       QueryTestCase(
           id=f"reject_{bt.value}_for_minLength_root",
           filter={"$jsonSchema": {"minLength": sample}},
           error_code=FAILED_TO_PARSE_ERROR,
           msg=f"minLength should reject {bt.value} at root",
       )
       for bt, sample in BSON_TYPE_SAMPLES.items()
       if bt not in {BsonType.INT, BsonType.LONG}
   ] + [
       QueryTestCase(
           id=f"reject_{bt.value}_for_exclusiveMinimum_root",
           filter={"$jsonSchema": {"exclusiveMinimum": sample, "minimum": 0}},
           error_code=TYPE_MISMATCH_ERROR,
           msg=f"exclusiveMinimum should reject {bt.value} at root",
       )
       for bt, sample in BSON_TYPE_SAMPLES.items()
       if bt not in {BsonType.BOOL}
   ]

   # --- Rejection nested in properties ---

   REJECTION_NESTED_TESTS: list[QueryTestCase] = [
       QueryTestCase(
           id=f"reject_{bt.value}_for_required_nested",
           filter={"$jsonSchema": {"properties": {"x": {"required": sample}}}},
           error_code=FAILED_TO_PARSE_ERROR if bt == BsonType.EMPTY_ARRAY else TYPE_MISMATCH_ERROR,
           msg=f"required should reject {bt.value} nested",
       )
       for bt, sample in BSON_TYPE_SAMPLES.items()
       if bt not in {BsonType.ARRAY}
   ] + [
       QueryTestCase(
           id=f"reject_{bt.value}_for_minLength_nested",
           filter={"$jsonSchema": {"properties": {"x": {"minLength": sample}}}},
           error_code=FAILED_TO_PARSE_ERROR,
           msg=f"minLength should reject {bt.value} nested",
       )
       for bt, sample in BSON_TYPE_SAMPLES.items()
       if bt not in {BsonType.INT, BsonType.LONG}
   ] + [
       QueryTestCase(
           id=f"reject_{bt.value}_for_exclusiveMinimum_nested",
           filter={
               "$jsonSchema": {"properties": {"x": {"exclusiveMinimum": sample, "minimum": 0}}}
           },
           error_code=TYPE_MISMATCH_ERROR,
           msg=f"exclusiveMinimum should reject {bt.value} nested",
       )
       for bt, sample in BSON_TYPE_SAMPLES.items()
       if bt not in {BsonType.BOOL}
   ]

   # --- Acceptance at root level ---

   ACCEPTANCE_ROOT_TESTS: list[QueryTestCase] = [
       QueryTestCase(
           id="accept_array_for_required_root",
           filter={"$jsonSchema": {"required": ["x"]}},
           doc=[{"_id": 1, "x": 1}],
           expected=[{"_id": 1, "x": 1}],
           msg="required should accept array at root",
       ),
   ] + [
       QueryTestCase(
           id=f"accept_{bt.value}_for_minLength_root",
           filter={"$jsonSchema": {"minLength": BSON_TYPE_SAMPLES[bt]}},
           doc=[{"_id": 1, "x": "hello"}],
           expected=[{"_id": 1, "x": "hello"}],
           msg=f"minLength should accept {bt.value} at root",
       )
       for bt in [BsonType.INT, BsonType.LONG]
   ] + [
       QueryTestCase(
           id="accept_bool_for_exclusiveMinimum_root",
           filter={"$jsonSchema": {"exclusiveMinimum": True, "minimum": 0}},
           doc=[{"_id": 1, "x": 1}],
           expected=[{"_id": 1, "x": 1}],
           msg="exclusiveMinimum should accept bool at root",
       ),
   ]

   # --- Acceptance nested in properties ---

   ACCEPTANCE_NESTED_TESTS: list[QueryTestCase] = [
       QueryTestCase(
           id="accept_array_for_required_nested",
           filter={"$jsonSchema": {"properties": {"x": {"required": ["x"]}}}},
           doc=[{"_id": 1, "x": 1}],
           expected=[{"_id": 1, "x": 1}],
           msg="required should accept array nested",
       ),
   ] + [
       QueryTestCase(
           id=f"accept_{bt.value}_for_minLength_nested",
           filter={"$jsonSchema": {"properties": {"x": {"minLength": BSON_TYPE_SAMPLES[bt]}}}},
           doc=[{"_id": 1, "x": "hello"}],
           expected=[{"_id": 1, "x": "hello"}],
           msg=f"minLength should accept {bt.value} nested",
       )
       for bt in [BsonType.INT, BsonType.LONG]
   ] + [
       QueryTestCase(
           id="accept_bool_for_exclusiveMinimum_nested",
           filter={"$jsonSchema": {"properties": {"x": {"exclusiveMinimum": True, "minimum": 0}}}},
           doc=[{"_id": 1, "x": 1}],
           expected=[{"_id": 1, "x": 1}],
           msg="exclusiveMinimum should accept bool nested",
       ),
   ]
 ```

danielfrankcom (commented):
Comments are on the BSON type enumeration approach, as discussed [here](https://github.com/documentdb/functional-tests/pull/152#pullrequestreview-4238040408). Have not reviewed the `$jsonSchema` coverage specifically.

documentdb_tests/compatibility/tests/core/operator/query/misc/jsonSchema/test_jsonSchema_keyword_validation.py
eerxuan: One reason that we do the extraction is to avoid missing a data type. So we should have a list of test case to go over with mechanically to ensure coverage. Your design shows the exact test case and value used, it's valuable for feature specific edge cases. If we are talking about the common argument test cases, I think it's okay that the actual tested value is extracted away. The PR design is also readable and cleaner, reader only need to know the intention of the test. After merge, the only time reader need to understand is when the test fails, and our log will print exact query used.

documentdb_tests/framework/test_constants.py
eerxuan: Yes, that slippery slope requires discussion. On one hand, we need to test those cases in most feature, it's good to abstract them. But in this PR, the approach only check if it passes. For other features, the index empty array is an important edge case, need to validate the result. I think for argument test, we just keep one case per type. Later if we want we can add more shared test cases, that would require a list of expected values.

eerxuan (approved):
I think this design is better for the argument testing. My updated design is still complex for argument type testing.

I think it's good idea to not have expected value list here, just verify pass. And separation of passing and failing cases is cleaner.

documentdb_tests/compatibility/tests/core/operator/query/misc/jsonSchema/test_jsonSchema_keyword_validation.py
danielfrankcom: > One reason that we do the extraction is to avoid missing a data type. So we should have a list of test case to go over with mechanically to ensure coverage. Your design shows the exact test case and value used [...]

Just want to clarify this since it may not be clear from the code snippet above, but the approach I included here _does_ use a list of test cases which is iterated mechanically to ensure coverage. The design above only includes exact values used for success cases, where there is only 1 accepted value.

For example the case below enumerates all types in `BSON_TYPE_SAMPLES`, except for `BOOL` which for this operator has a different expected value (it's accepted).

```python
QueryTestCase(
    id=f"reject_{bt.value}_for_exclusiveMinimum_root",
    filter={"$jsonSchema": {"exclusiveMinimum": sample, "minimum": 0}},
    error_code=TYPE_MISMATCH_ERROR,
    msg=f"exclusiveMinimum should reject {bt.value} at root",
)
for bt, sample in BSON_TYPE_SAMPLES.items()
if bt not in {BsonType.BOOL}
```

documentdb_tests/framework/test_constants.py
vic-tsang: okay I can make the change to default these types with: 
BsonType.OBJECT: {}
~~BsonType.ARRAY: []~~

~~I don't think the test actually needs any value in them as we don't verify the value of the result.~~

Update: Actually we cant merge the ARRAY and EMPTY_ARRAY into one type as we will be skipping over how they are tested. They are both the same type but has different results ARRAY is valid while EMPTY_ARRAY is Invalid. If we set ARRAY as empty ARRAY and put it in the valid list. The test will never test ARRAY as empty as it is a valid type. We can only change OBJECT to empty that's fine.

vic-tsang: as to the data types using the constants I thought I might as well use an existing value rather than defining a new one. It's not my intention to test edge cases here. This test is strictly for checking if the BSON type is accepted or not regardless of the value.

eerxuan: > Update: Actually we cant merge the ARRAY and EMPTY_ARRAY into one type as we will be skipping over how they are tested. They are both the same type but has different results ARRAY is valid while EMPTY_ARRAY is Invalid. If we set ARRAY as empty ARRAY and put it in the valid list. The test will never test ARRAY as empty as it is a valid type. We can only change OBJECT to empty that's fine.

I you can keep non-empty array and object, they will be shared. If you need to test empty array, add as separate edge case for the feature.

documentdb_tests/compatibility/tests/core/operator/query/misc/jsonSchema/test_jsonSchema_keyword_validation.py
eerxuan: Okay, but there's no guarantee that the type in exception list is covered in another for TestCase. We still have multiple isolated test cases.

karthikvarma22 (approved):

documentdb_tests/compatibility/tests/core/operator/query/misc/jsonSchema/test_jsonSchema_keyword_validation.py
danielfrankcom: I mentioned this in more detail in my comment [here](https://github.com/documentdb/functional-tests/pull/152#pullrequestreview-4238040408), but if an operator accepts a value then we should already have extensive testing of the success cases for that type.

If we cover both accept/reject cases with this approach then even though every BSON type is covered, our coverage is extremely weak for the accepted value. My preference would be to simplify the test definition so it's clear what's actually being tested, and call out that we're excluding a type which should be well-tested elsewhere. The alternative in my mind is we cover the type which we're missing extensive tests for in the loop of BSON types, but effectively hide the fact that it's tested in a weak way.

With the pattern we are establishing for `BsonTypeTestCase` I think we will end up with more code for most operators, and bespoke test methods for accepted/rejected values. Instead, we could have `QueryTestCase` or equivalent which fits into the data-driven list of tests which are already widely used in the test files.

I won't block this change but I'm still not convinced we need to add `BsonTypeTestCase` and surrounding machinery for the 99% usecase which is quite simple and excludes 1 or 2 types which are already well tested. Though it has more applicability for `jsonSchema`, even here it seems the same thing could be accomplished without the extra machinery using `QueryTestCase` which also makes it clearer what is being tested at the callsite imo.

eerxuan (approved):

documentdb_tests/compatibility/tests/core/operator/query/misc/jsonSchema/test_jsonSchema_keyword_validation.py
eerxuan: Okay, will merge this PR and have separate discussion on which format to use for other features.

documentdb_tests/framework/test_constants.py
vic-tsang: I didn't add the customized BsonType list (empty object, empty array)  into jsonSchema test. It was over complicated the framework just to support a custom list. If we need to test edge cases, we should just create a test case outside of the framework.

eerxuan (approved):

PR #170, opened 2026-05-07
Add $planCacheStats stage tests

danielfrankcom:
This change adds test coverage for the `$planCacheStats` aggregation stage operator.

As part of this change I added `target_collection` and `indexes` to `StageTestCase` as I found I was needing them more than originally expected, and having to create too many custom `StageTestCase` derivations. For tests that don't need them they are optional, so this shouldn't cause any problems. They are also generally applicable, and while they may not be useful for many aggregation stage tests, they at least make sense still in the context of those stages.

As with tests like #166, I've used `assertProperties` to verify what I can about the expected output shape for this operator, but it is highly dependent on database engine internals. The main thing I focused on testing was the input/output spec, without trying to overreach and verify functionality of the operator.

Ref: #34


documentdb_tests/compatibility/tests/core/operator/stages/planCacheStats/test_planCacheStats_output.py
```diff
@@ -0,0 +1,122 @@
...
+    StageTestCase(
+        "collection_type_view",
+        target_collection=ViewCollection(),
+        docs=[],
+        pipeline=[{"$planCacheStats": {}}],
+        expected=[],
+        msg="$planCacheStats on view should succeed",
+    ),
+]
+
```
SaiPradyumn, line 116: can we have a happy-path test that confirms {"allHosts": false} succeeds and returns the same results as {}

documentdb_tests/compatibility/tests/core/operator/stages/planCacheStats/test_planCacheStats_output.py
```diff
@@ -0,0 +1,122 @@
...
+        docs=[{"_id": 1, "value": 10}],
+        pipeline=[{"$planCacheStats": {}}],
+        expected=[],
+        msg=(
+            "$planCacheStats on a collection with documents but no prior"
+            " queries should return empty"
+        ),
+    ),
+]
+
```
SaiPradyumn, line 50: can we add a test where we Warm the cache, clear it, then confirm $planCacheStats returns empty again. This is to validate the lifecycle behavior.

documentdb_tests/compatibility/tests/core/operator/stages/planCacheStats/test_planCacheStats_output.py
```diff
@@ -0,0 +1,122 @@
...
+            "cachedPlan": IsType("object"),
+            "shard": NotExists(),
+            "queryHash": IsType("string"),
+            "creationExecStats": Exists(),
+            "candidatePlanScores": Exists(),
+        },
+        msg="$planCacheStats output document structure",
+    ),
+]
+
```
SaiPradyumn, line 95: The warmed-cache tests run a single query shape. No test verifies behavior with multiple distinct
  query shapes producing multiple cache entries (e.g., c.find({"a": 5}) AND c.find({"b": 3}) → verify count >= 2).

documentdb_tests/compatibility/tests/core/operator/stages/planCacheStats/test_planCacheStats_output.py
danielfrankcom: The tests here are intended to cover to the input/output spec of `$planCacheStats` itself. It sounds like this kind of test would lean towards testing engine functionality rather than the stage's output contract, which I think we had decided we wouldn't attempt to cover.

SaiPradyumn (approved):

eerxuan (approved):

PR #172, opened 2026-05-08
Add $compact command tests

danielfrankcom:
This change adds tests for the `$compact` collection command. As with other commands, the intent is to test the input/output spec rather than the actual compaction functionality of the underlying engine.

As part of this change I removed `commands/utils/target_collection.py` as I noticed it was still hanging around. This was originally intended to be moved to the `framework/target_collection.py` file, but it seems a new file was created without deleting the original version in the `commands` dir.

Ref: #26


documentdb_tests/compatibility/tests/core/collections/commands/compact/test_compact_freespace.py
```diff
@@ -0,0 +1,266 @@
+"""Tests for compact command freeSpaceTargetMB validation."""
+
```
SaiPradyumn, line 2: No happy-path test exists for this parameter. Every test is an error case. We never confirm it actually works

documentdb_tests/compatibility/tests/core/collections/commands/compact/test_compact_maxtimems.py
```diff
@@ -0,0 +1,290 @@
...
+        docs=[{"_id": 1}],
+        command=lambda ctx: {
+            "compact": ctx.collection,
+            "maxTimeMS": DECIMAL128_NEGATIVE_ZERO,
+        },
+        expected={"bytesFreed": 0, "ok": 1.0},
+        msg="maxTimeMS=Decimal128(-0) should be accepted",
+    ),
+]
+
```
SaiPradyumn, line 124: The maxTimeMS file tests numeric edge cases but doesn't test non-numeric BSON types (string, bool, array, object, ObjectId, etc.) which should produce TYPE_MISMATCH_ERROR. The listDatabases PR had these explicitly.

documentdb_tests/compatibility/tests/core/collections/commands/compact/test_compact_freespace.py
danielfrankcom: There are happy path tests in one of the other files but they have mixed purpose, hence their exclusion from this file. I've added some dedicated ones just to prove out this functionality within the file.

SaiPradyumn (approved):

eerxuan (approved):

Issue #27, opened 2026-04-03
Add compatibility test for core: indexes (14 total)


eerxuan:
core	indexes	commands	-	-	createIndexes
core	indexes	commands	-	-	dropIndexes
core	indexes	commands	-	-	listIndexes
core	indexes	commands	-	-	reIndex
core	indexes	properties	-	-	case-insensitive
core	indexes	properties	-	-	partial
core	indexes	properties	-	-	sparse
core	indexes	properties	-	-	ttl
core	indexes	properties	-	-	unique
core	indexes	types	-	-	compound
core	indexes	types	-	-	geospatial
core	indexes	types	-	-	multikey
core	indexes	types	-	-	single
core	indexes	types	-	-	text

PR #177, opened 2026-05-11
Added core indexes property tests for ttl

vic-tsang:
This PR contains:
- core indexes property tests for ttl

Ref
- Issue  #27


karthikvarma22 (approved):

documentdb_tests/compatibility/tests/core/indexes/properties/ttl/test_ttl_create.py
```diff
@@ -0,0 +1,134 @@
...
+            {
+                "key": {"d": 1},
+                "name": "ttl_nz_dec",
+                "expireAfterSeconds": DECIMAL128_NEGATIVE_ZERO,
+            },
+        ),
+        msg="Should accept negative zero Decimal128 (stored as 0)",
+    ),
+    IndexTestCase(
+        id="value_fractional_double",
```
eerxuan: 🟡 [Suggestion] value_fractional_double / value_fractional_decimal128 only test acceptance, not the
  truncation behavior

  IndexTestCase(
      id="value_fractional_double",
      indexes=({"key": {"d": 1}, "name": "ttl_frac_d", "expireAfterSeconds": 3600.5},),
      msg="Should accept fractional double (truncated to integer)",
  ),

  The msg says "truncated to integer" but the test only verifies acceptance. To prove the truncation,
  follow with a listIndexes and assert expireAfterSeconds == 3600 (integer, not 3600.5). This is the kind
  of behavior verification the project values per the spec-level testing approach.


This kind of specs need a verification on the actual value.

vic-tsang: verified truncation now.

eerxuan (approved):

PR #178, opened 2026-05-12
Added core indexes property test for unique

vic-tsang:
This PR contains 
- core indexes property test for unique

Ref
- Issue #27


documentdb_tests/compatibility/tests/core/indexes/properties/unique/test_unique_constraint.py
karthikvarma22, line 1: "invalid_input", is this a valid naming convention ?

documentdb_tests/compatibility/tests/core/indexes/properties/unique/test_unique_errors.py
karthikvarma22, line 1: can we also have an empty array [] duplicate case ?

vic-tsang: added

documentdb_tests/compatibility/tests/core/indexes/properties/unique/test_unique_constraint.py
vic-tsang: This should only be used for the error case. For the happy case, we should use the newly added input.

karthikvarma22 (approved):

documentdb_tests/compatibility/tests/core/indexes/properties/unique/test_unique_errors.py
```diff
@@ -0,0 +1,542 @@
...
+    result = execute_command(
+        collection,
+        {"insert": collection.name, "documents": [{"v": 2}, {"v": 1}, {"v": 3}], "ordered": True},
+    )
+    errors = [{"index": e["index"], "code": e["code"]} for e in result["writeErrors"]]
+    assertSuccessPartial(
+        {"ok": 1.0, "errors": errors},
+        {"ok": 1.0, "errors": [{"index": 1, "code": DUPLICATE_KEY_ERROR}]},
+        msg="Ordered batch should have one writeError at index 1 with DuplicateKey code",
+    )
```
eerxuan: Why we don't assert on result directly?

documentdb_tests/compatibility/tests/core/indexes/properties/unique/test_unique_errors.py
```diff
@@ -0,0 +1,542 @@
...
+        },
+    )
+    assertFailureCode(
+        result,
+        DUPLICATE_KEY_ERROR,
+        msg="Should fail to create unique index with existing duplicates",
+    )
+
+
+def test_unique_id_index_cannot_be_dropped(collection):
```
eerxuan: 🟡 [Suggestion] test_unique_id_index_cannot_be_dropped is a dropIndexes test, not a unique test

vic-tsang: removed

documentdb_tests/compatibility/tests/core/indexes/properties/unique/test_unique_errors.py
vic-tsang: actually we did it in `test_unique_unordered_batch_continues_past_duplicates`. In this case `test_unique_unordered_batch_reports_all_duplicates`, the test is trying to assert all the errors with the same input. I am guessing why they splitted into two test cases is because we cant assert twice in one test case. 

The intention for the test cases are a little bit different 
`test_unique_unordered_batch_continues_past_duplicates` - proves that the server didnt give up and it still inserted the valid document 
`test_unique_unordered_batch_reports_all_duplicates` - proves that all failures are returned not just the first one

eerxuan: I mean why we construct a fake response {"ok": 1.0, "errors": errors}, and the "ok": 1.0 has no meaning here.

vic-tsang: you are correct. I have added nested structure comparison in assertPartialSuccess so now it's able to assert without creating an artificial result.

eerxuan (approved):

PR #181, opened 2026-05-13
Add $validateDBMetadata command tests

danielfrankcom:
This change adds tests for the `$validateDBMetadata` collection command.


documentdb_tests/compatibility/tests/core/collections/commands/validateDBMetadata/test_validateDBMetadata_api_version.py
```diff
@@ -0,0 +1,210 @@
+"""Tests for validateDBMetadata API version validation behavior."""
+
```
SaiPradyumn, line 2: deprecationErrors: true / false behavior  : This is a documented required parameter with distinct behavior from strict. Currently untested entirely.  It's a core feature of the command.

documentdb_tests/compatibility/tests/core/collections/commands/validateDBMetadata/test_validateDBMetadata_index_types.py
```diff
@@ -0,0 +1,357 @@
...
+
+from documentdb_tests.compatibility.tests.core.collections.commands.utils.command_test_case import (
+    CommandContext,
+    CommandTestCase,
+)
+from documentdb_tests.framework.assertions import assertResult
+from documentdb_tests.framework.error_codes import API_STRICT_ERROR
+from documentdb_tests.framework.executor import execute_admin_command
+from documentdb_tests.framework.parametrize import pytest_params
+
```
SaiPradyumn, line 16: strict: false ->  Verifies the negative case: that the parameter actually controls output. Without it, we only know strict: true produces errors, not that false suppresses them.

danielfrankcom: This should already be covered in `test_validateDBMetadata_api_version.py` by `api_version_1_strict_false` and `api_version_1_strict_false_sparse`.

documentdb_tests/compatibility/tests/core/collections/commands/validateDBMetadata/test_validateDBMetadata_api_version.py
danielfrankcom: There are some existing tests for `deprecationErrors` in this file, including cases for `strict: true` + `deprecationErrors: true`, `strict: false` + `deprecationErrors: true`, and both omitted. The type strictness file also validates that non-boolean values are rejected. That said, the `strict: true, deprecationErrors: false` case to prove that `false` doesn't suppress strict errors was missing, and I've added it.

The reason there isn't broader coverage of the outputs is they are inherently engine-specific and not part of the compatibility service we are aiming to test here. They would also not be stable, and we would have to update what we were using over time as versions change. In the current MongoDB version for example, `deprecationErrors` has no observable effect for this command. All valid index types and view pipelines with `strict: false, deprecationErrors: true` produce no errors. The one historical candidate that might have (`geoHaystack`) was removed in 5.0 and can't be created anymore. A different compatible engine might have a different set of deprecated components.

The tests capture the actual behavior as best we can - `deprecationErrors` is accepted, type-checked, and currently a no-op for `validateDBMetadata`. There's nothing I can find that triggers deprecation
 warnings without also triggering strict errors. If you know of something I'm missing I'm happy to add it.

SaiPradyumn (approved):

eerxuan (approved):

eerxuan (approved):

PR #185, opened 2026-05-14
Added core indexes property tests for sparse

vic-tsang:
This PR contains:
- core indexes property tests for sparse

Ref:
- Issue #27


documentdb_tests/compatibility/tests/core/indexes/properties/sparse/test_sparse_crud.py
```diff
@@ -0,0 +1,199 @@
...
+
+pytestmark = pytest.mark.index
+
+FAM_DOCS = (
+    {"_id": 1, "a": 1},
+    {"_id": 2, "a": 2},
+    {"_id": 3},
+)
+
+SPARSE_CRUD_COUNT_TESTS: list[IndexTestCase] = [
```
vic-tsang: @karthikvarma22 I am not exactly sure if findAndModify and update command should be tested in this folder. Please let me know if it doesnt belong.

vic-tsang: Confirmed with Yunxuan these might not be applicable as they are more functional tests. I will remove all update/modify related tests

documentdb_tests/compatibility/tests/core/indexes/properties/test_properties_combinations.py
```diff
@@ -0,0 +1,245 @@
...
+            {
+                "key": {"dateField": 1},
+                "name": "ttl_sparse",
+                "expireAfterSeconds": 3600,
+                "sparse": True,
+            },
+        ),
+        msg="Should allow TTL + sparse",
+    ),
+    IndexTestCase(
```
karthikvarma22, line 40: can we have sparse + partialFilterExpression test ?

documentdb_tests/compatibility/tests/core/indexes/properties/sparse/test_sparse_create.py
```diff
@@ -0,0 +1,129 @@
...
+        id="descending",
+        indexes=({"key": {"a": -1}, "name": "idx_desc_sparse", "sparse": True},),
+        msg="Should create sparse index on descending field",
+    ),
+    IndexTestCase(
+        id="hashed",
+        indexes=({"key": {"a": "hashed"}, "name": "idx_hashed_sparse", "sparse": True},),
+        msg="Should create sparse index on hashed field",
+    ),
+    IndexTestCase(
```
karthikvarma22, line 41: index_created_response() why not use this for all cases here ?

documentdb_tests/compatibility/tests/core/indexes/properties/test_properties_combinations.py
vic-tsang: added

documentdb_tests/compatibility/tests/core/indexes/properties/sparse/test_sparse_create.py
vic-tsang: in this case the response is 
`{"ok": 1.0, "createdCollectionAutomatically":
  true, "numIndexesAfter": M}
` 
 Note: numIndexesBefore is missing in the response. It seems like a one off case which does this. Hence we didnt use index_created_response.

karthikvarma22 (approved):

documentdb_tests/compatibility/tests/core/indexes/properties/sparse/test_sparse_index_coverage.py
```diff
@@ -0,0 +1,230 @@
...
+    )
+    collection.insert_many(list(test.doc))
+    result = execute_command(
+        collection,
+        {"count": collection.name, "query": {}, "hint": test.command_options["hint"]},
+    )
+    assertSuccess(result, test.expected, raw_res=True)
+
+
+def test_sparse_text_skips_missing_text(collection):
```
eerxuan: 🟠 [Required] test_sparse_text_skips_missing_text and test_sparse_exists_false_not_affected both belong outside properties/sparse/

  def test_sparse_text_skips_missing_text(collection):
      """Test text index is always sparse — skips docs without text field."""
      ...indexes=[{"key": {"content": "text"}, "name": "idx_text"}],

  This test creates a text index without sparse: True and asserts text-index-implicit-sparse behavior. The "text index is always sparse" property belongs in a text-index test folder (or
   properties/sparse/test_implicit_sparse.py), not in this folder which is about explicit sparse option.

  Same for test_sparse_exists_false_not_affected — the property tested is "$exists:false works correctly" which is a $exists query operator concern, not sparse-specific.

  These two would live better in:
  - text_search/ for the text-implicit-sparse test
  - query/misc/exists/ for the $exists:false test

vic-tsang: removed

eerxuan (approved):

PR #187, opened 2026-05-15
Add $renameCollection command tests

danielfrankcom:
This change adds tests for the `$renameCollection` command.


SaiPradyumn (commented):
writeConcern parameter : do we need a separate file for this test? its a documented feature.

documentdb_tests/compatibility/tests/core/collections/commands/renameCollection/test_renameCollection_special_collections.py
```diff
@@ -0,0 +1,336 @@
...
+from documentdb_tests.framework.executor import execute_admin_command
+from documentdb_tests.framework.parametrize import pytest_params
+from documentdb_tests.framework.target_collection import (
+    CappedCollection,
+    ExistingCollection,
+    SiblingCollection,
+    TimeseriesCollection,
+    ViewCollection,
+)
+
```
SaiPradyumn, line 28: Source is a view : Documented restriction ("views cannot be renamed"). If untested, a real  compatibility gap.

similarly for Source is a timeseries.

documentdb_tests/compatibility/tests/core/collections/commands/renameCollection/test_renameCollection_null_missing.py
```diff
@@ -0,0 +1,148 @@
+"""Tests for renameCollection null and missing field behavior."""
```
SaiPradyumn, line 1: to field type errors  (int, array, etc.)

The null_missing file only tests to: null and to omitted (both → MISSING_FIELD_ERROR). The namespace_format file only uses string values. No test passes non-string types like "to": 123 or "to": [].

documentdb_tests/compatibility/tests/core/collections/commands/renameCollection/test_renameCollection_null_missing.py
```diff
@@ -0,0 +1,148 @@
...
+
```
SaiPradyumn, line 2: renameCollection (Same situation — null_missing tests renameCollection): null (→ MISSING_FIELD_ERROR), but no test passes int/array/object/bool to the renameCollection field.

documentdb_tests/compatibility/tests/core/collections/commands/renameCollection/test_renameCollection_special_collections.py
danielfrankcom: This should be covered by `rename_view_fails` and `rename_timeseries_fails`

documentdb_tests/compatibility/tests/core/collections/commands/renameCollection/test_renameCollection_null_missing.py
danielfrankcom: This should be covered in the `test_renameCollection_type_strictness_rename_to.py` file

documentdb_tests/compatibility/tests/core/collections/commands/renameCollection/test_renameCollection_null_missing.py
danielfrankcom: This should be covered in the `test_renameCollection_type_strictness_rename_to.py` file

SaiPradyumn (approved):

eerxuan (approved):

Issue #32, opened 2026-04-03
Add compatibility test for query: geospatial operators (11 total)


eerxuan:
core	operator	query	geospatial	-	geoIntersects
core	operator	query	geospatial	-	geoWithin
core	operator	query	geospatial	-	near
core	operator	query	geospatial	-	nearSphere
core	operator	query	geospatial	specifiers	box
core	operator	query	geospatial	specifiers	center
core	operator	query	geospatial	specifiers	centerSphere
core	operator	query	geospatial	specifiers	geometry
core	operator	query	geospatial	specifiers	maxDistance
core	operator	query	geospatial	specifiers	minDistance
core	operator	query	geospatial	specifiers	polygon

PR #192, opened 2026-05-15
Add $polygon tests

PatersonProjects:
The PR adds the tests for the $polygon geospatial specifier, and one additional error case test for geoWithin generally

Issue: Ref #32


vic-tsang (changes requested):
Edge Cases: 

- Point located exactly on a polygon vertex
- Point collinear with an edge but beyond the segment
- Point on a non-axis-aligned edge (e.g., midpoint of hypotenuse) 

Errors: 
 - `$polygon: [[0,0,0],[1,1,1],[2,0,0]]` | 3-element points — error or silently ignored? 
 - `$polygon: [[], [], []]` | Empty sub-arrays as points 
-  Multiple shape specifiers in one $geoWithin
- `$polygon: [[0,0],[1,1],[NaN,NaN],[2,0]]` | Error even if most points are valid 
- `[[-Inf, 0], [1,1], [2,0]]` | Only positive Infinity tested 

Others: 
- embedded_doc_non_xy_keys | {"a":1, "b":2} — does it use first two fields as x/y?  Behavior is non-obvious
= a test case with $not + $polygon

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_index_interaction.py
```diff
@@ -0,0 +1,106 @@
...
+    result = execute_command(
+        collection,
+        {
+            "find": collection.name,
+            "filter": {"loc": {"$geoWithin": {"$polygon": [[4, 4], [6, 4], [5, 6]]}}},
+        },
+    )
+    # Verify it returns results without error (exact count depends on boundary inclusion)
+    result_docs = result["cursor"]["firstBatch"]
+    assertSuccess(result, result_docs, msg="Triangle on dense grid should not error")
```
eerxuan: This assertion is dead weight, it will error out on line 105 if failed. 
> exact count depends on boundary inclusion
why don't we assert the exact result set?

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_query_interaction.py
```diff
@@ -0,0 +1,261 @@
...
+            "find": collection.name,
+            "filter": {"loc": {"$geoWithin": {"$polygon": [[0, 0], [0, 10], [10, 10], [10, 0]]}}},
+            "limit": 2,
+        },
+    )
+    # Should return exactly 2 documents
+    result_docs = result["cursor"]["firstBatch"]
+    assertSuccess(
+        result, result_docs[:2], raw_res=False, msg="$polygon with limit should limit results"
+    )
```
eerxuan: this is asserting to itself

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_query_interaction.py
```diff
@@ -0,0 +1,261 @@
...
+        filter={
+            "$expr": {
+                "$eq": [
+                    {
+                        "$literal": {
+                            "loc": {
+                                "$geoWithin": {"$polygon": [[0, 0], [0, 10], [10, 10], [10, 0]]}
+                            }
+                        }
+                    },
```
eerxuan: tests the wrong thing

  The test wraps the geo query in $literal and compares to True:
  "$expr": {"$eq": [{"$literal": {"loc": {"$geoWithin": ...}}}, True]}
  This is testing {some object} != True, which is trivially false → empty result. The PR's msg says
  "$polygon inside $expr should not match (not evaluated as geo query)" — but the actual behavior of
  $geoWithin directly inside $expr is error code 168 "Unrecognized expression '$geoWithin'". Two fixes
  worth considering:
  - Replace with assertFailureCode(result, 168, ...) and filter {"$expr": {"$geoWithin": ...}} directly,
  which actually exercises the non-support.
  - Or remove the test — wrapping in $literal proves nothing about $polygon specifically.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_query_interaction.py
```diff
@@ -0,0 +1,261 @@
...
+        },
+    )
+    # Should return exactly 2 documents
+    result_docs = result["cursor"]["firstBatch"]
+    assertSuccess(
+        result, result_docs[:2], raw_res=False, msg="$polygon with limit should limit results"
+    )
+
+
+def test_polygon_on_capped_collection(database_client):
```
eerxuan: It's better put under capped collection.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_index_interaction.py
PatersonProjects: Good catch, I've reworked the test to have an inline expected set of points asserted against the actual results, rather than just confirming no error.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_edge_cases.py
```diff
@@ -0,0 +1,194 @@
...
+    ),
+    QueryTestCase(
+        id="consecutive_duplicate_points",
+        filter={"loc": {"$geoWithin": {"$polygon": [[0, 0], [0, 0], [5, 5], [10, 0]]}}},
+        doc=[{"_id": 1, "loc": [4, 2]}, {"_id": 2, "loc": [20, 20]}],
+        expected=[{"_id": 1, "loc": [4, 2]}],
+        msg="Polygon with consecutive duplicate points should still work",
+    ),
+    QueryTestCase(
+        id="self_intersecting_bowtie",
```
eerxuan, line 63: Add another shape:
  # 5-pointed star: classic single-ring self-intersecting polygon.
  # The geometric center has winding number 2, which under even-odd parity is OUT.
  QueryTestCase(
      id="five_pointed_star",
      filter={"loc": {"$geoWithin": {"$polygon": [
          [5.0, 10.0], [7.939, 0.955], [0.245, 6.545],
          [9.755, 6.545], [2.061, 0.955]
      ]}}},
      doc=[
          {"_id":1,"loc":[5, 5]},   # geometric center
          {"_id":2,"loc":[5, 9]},   # in upper point
          {"_id":3,"loc":[15,15]},  # outside
      ],
      expected=[{"_id":2,"loc":[5,9]}],  # only the upper point matches
      msg="Five-pointed star: even-odd parity excludes the geometric center",
  ),

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_edge_cases.py
```diff
@@ -0,0 +1,194 @@
...
+                    "$polygon": [[0.0001, 0.0001], [0.0001, 0.001], [0.001, 0.001], [0.001, 0.0001]]
+                }
+            }
+        },
+        doc=[{"_id": 1, "loc": [0.0005, 0.0005]}, {"_id": 2, "loc": [1, 1]}],
+        expected=[{"_id": 1, "loc": [0.0005, 0.0005]}],
+        msg="Polygon with very small fractional coordinates should work",
+    ),
+    QueryTestCase(
+        id="longitude_latitude_bounds",
```
eerxuan: 2d shape should not have bounds, change this test case to a large value test. The bounds only exist for 2dsphere shapes in GeoJson format.

PatersonProjects: Removed test as large bounds already tested in core_functionality

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_edge_cases.py
PatersonProjects: Good test case, added!

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_query_interaction.py
PatersonProjects: Removed as out of scope for polygon.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_query_interaction.py
PatersonProjects: Removed test as out of scope for polygon, decided to go with that option as expr already has tests for error code 168.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/polygon/test_polygon_query_interaction.py
PatersonProjects: Fixed, added sort as well to ensure expected would always match, may be unnecessary due to id numbering but figured it was the safer implementation, let me know if you would prefer it reverted to just limit.

eerxuan (approved):
Thanks for finding the smoke test issue, can you raise another PR to fix it?

PR #193, opened 2026-05-15
Added geospatial tests for $center

vic-tsang:
This PR contains:
- geospatial tests for $center

Ref:
- Issue #32


karthikvarma22 (approved):

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/center/test_center_query_interaction.py
```diff
@@ -0,0 +1,239 @@
...
+TESTS: list[QueryTestCase] = [
+    QueryTestCase(
+        id="two_geoWithin_same_field",
+        filter={
+            "$and": [
+                {"loc": {"$geoWithin": {"$center": [[0, 0], 5]}}},
+                {"loc": {"$geoWithin": {"$center": [[1, 1], 5]}}},
+            ]
+        },
+        doc=[{"_id": 1, "loc": [0, 0]}],
```
eerxuan: only inserts one document {_id:1,
  loc:[0,0]}. The test passes, but it doesn't actually demonstrate the AND filtering (a doc only in one
  circle would also need to be present and excluded). Suggest adding {_id:2, loc:[3,3]} (in second circle
  only) and {_id:3, loc:[10,10]} (in neither), and asserting only _id:1 returns. Otherwise the assertion
  is non-discriminating.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/center/test_center_index_behavior.py
```diff
@@ -0,0 +1,209 @@
...
+        ignore_doc_order=True,
+        msg="Should handle center outside 2d index bounds",
+    )
+
+
+def test_center_2d_index_mixed_legacy_and_geojson(collection):
+    """Test $center with 2d index on collection with both legacy and GeoJSON documents."""
+    collection.create_index([("loc", "2d")])
+    collection.insert_many(
+        [
```
eerxuan: name says "mixed
  legacy and GeoJSON" but the inserted docs ([5,5] and [50,50]) are both legacy. The original mixed-corpus
   intent is missing. Either add a GeoJSON Point doc or rename the test to drop
  "mixed_legacy_and_geojson".

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/center/test_center_query_interaction.py
vic-tsang: revised test

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/center/test_center_index_behavior.py
vic-tsang: actually this test is invalid as we cant add GeoJSON to a collection with 2d index. I think that's why it didnt want to add the GeoJSON otherwise it will error. Remove this test case

eerxuan (approved):

PR #191, opened 2026-05-15
Add `count` aggregation command tests

danielfrankcom:
This change adds tests for the `count` aggregation command.

PR #215, opened 2026-05-19
Add `distinct` aggregation command tests

danielfrankcom:
This change adds tests for the `distinct` aggregation command.

There is some overlap here with #191 as some of the framework changes are needed by both. I've included them here so they should merge cleanly.


eerxuan (approved):
Other than view and collation trim as discussed before, LGTM.

eerxuan (approved):

PR #218, opened 2026-05-20
Added geospatial tests for $near

vic-tsang:
This PR contains:
- geospatial tests for $near
- added conftest.py under geospatial to enable index creation with 2dsphere and 2d to be used with nearSphere later 

Ref:
- Issue #32


documentdb_tests/compatibility/tests/core/operator/query/geospatial/near/test_near_core_functionality.py
```diff
@@ -0,0 +1,113 @@
+"""Tests for $near core functionality — nearest-first sorting, query interactions, nested fields."""
```
fanyangv, line 1: Missing explicit sort override test. The MongoDB docs state: "If you use the sort() method in your query, MongoDB performs a second sort operation, re-ordering the matching documents." PR #222 has this for $nearSphere (test_nearSphere_explicit_sort_overrides_distance) — please add an equivalent here for $near that combines $near with an explicit "sort" in the find command and verifies the distance ordering is overridden.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/near/test_near_errors.py
```diff
@@ -0,0 +1,350 @@
...
+            ],
+            "cursor": {},
+        },
+    )
+    assertFailureCode(result, NEAR_NOT_ALLOWED_ERROR, msg="Should reject $near in aggregate $match")
+
+
+def test_near_combined_with_text_errors(collection):
+    """Verifies $near cannot be combined with $text in the same query."""
+    collection.create_index([("loc", "2dsphere")])
```
fanyangv: This create_index on loc is redundant — the module-level `pytestmark =  pytest.mark.usefixtures("geo_2dsphere")` already creates a 2dsphere index on loc for every test in this file. Only the text index line is needed here.

fanyangv (changes requested):

documentdb_tests/compatibility/tests/core/operator/query/geospatial/near/test_near_core_functionality.py
vic-tsang: added

documentdb_tests/compatibility/tests/core/operator/query/geospatial/near/test_near_errors.py
vic-tsang: removed

fanyangv (approved):

eerxuan (approved):

documentdb_tests/compatibility/tests/core/operator/query/geospatial/conftest.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+
+
+@pytest.fixture
+def geo_2dsphere(collection):
```
eerxuan, line 5: I missed this, remember to keep each test case self-contained: has setup command and expected. So 
```
pytestmark = pytest.mark.usefixtures("geo_2dsphere")
```
file level fixture breaks the pattern.

Also, you have fixture and inline create_index, which means this fixture is not flexible enough. 

Please remove the fixture. We should have consistent format for index setup and collection setup. 
We can create a similar index helpers as the CustomCollection etc. This can keep the self-contained test case pattern, and it is flexible enough so we can have TestCase specific index setup.
```
    CommandTestCase(
        "collation_null_uses_default",
        target_collection=CustomCollection(options={"collation": {"locale": "en", "strength": 2}}),
        docs=[{"_id": 1, "s": "abc"}, {"_id": 2, "s": "ABC"}, {"_id": 3, "s": "def"}],
        command=lambda ctx: {
            "count": ctx.collection,
            "query": {"s": "abc"},
            "collation": None,
        },
        expected={"n": 2, "ok": 1.0},
        msg="count with collation=null should use collection default collation",
    ),
```

vic-tsang: Addressed in #249

PR #220, opened 2026-05-20
Add $geometry tests

PatersonProjects:
This PR adds the compatibility test for the $geometry geospatial specifier

Ref: Issue #32


documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_index_behavior.py
```diff
@@ -0,0 +1,84 @@
...
+
+@pytest.mark.parametrize("test", pytest_params(INDEX_REQUIRED_TESTS))
+def test_geometry_index_required_errors(collection, test):
+    """Verifies $near with $geometry errors without a 2dsphere index."""
+    collection.insert_many(test.doc)
+    result = execute_command(collection, {"find": collection.name, "filter": test.filter})
+    assertFailureCode(result, test.error_code, msg=test.msg)
+
+
+@pytest.mark.parametrize("test", pytest_params(NO_INDEX_SUCCESS_TESTS))
```
vic-tsang: I suggest grouping the errors into the test_geometry_errors.py

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_argument_handling.py
```diff
@@ -0,0 +1,126 @@
...
+from documentdb_tests.framework.assertions import assertSuccess
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+
+STRICT_CRS = {
+    "type": "name",
+    "properties": {"name": "urn:x-mongodb:crs:strictwinding:EPSG:4326"},
+}
+
+# --- Success cases: valid $geometry structures ---
```
vic-tsang: I suggest we remove this doesn't really add much value

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_argument_handling.py
```diff
@@ -0,0 +1,126 @@
...
+        ],
+        expected=[{"_id": 1, "loc": {"type": "Point", "coordinates": [0.5, 0.5]}}],
+        msg="Should accept custom CRS with $geoWithin",
+    ),
+    QueryTestCase(
+        id="array_value_no_error",
+        filter={"loc": {"$geoIntersects": {"$geometry": [0, 0]}}},
+        doc=[{"_id": 1, "loc": {"type": "Point", "coordinates": [0, 0]}}],
+        expected=[{"_id": 1, "loc": {"type": "Point", "coordinates": [0, 0]}}],
+        msg="$geometry: array value does not error, geometry specifier is silently ignored",
```
vic-tsang: The original test had an incorrect msg claiming $geometry: [0, 0] is "silently ignored." In reality, MongoDB treats the array as a legacy coordinate pair and performs a real geospatial match. The test only passed by coincidence (single doc
  that happened to match). Fixed by adding a second non-matching doc and correcting the msg.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_argument_handling.py
```diff
@@ -0,0 +1,126 @@
+"""Tests for $geometry valid argument handling —
```
vic-tsang, line 1: consider adding these 

  1. large_polygon_greater_than_hemisphere — verifies MongoDB auto-corrects to smaller complement area
  2. lng_180_does_not_match_negative_180 — verifies -180 and 180 are NOT treated as equivalent (strict numeric)
  3. north_pole_any_longitude — verifies [0,90] matches [180,90] (longitude irrelevant at poles)
  4. string_loc_insert_rejected — verifies 2dsphere index rejects insert of non-geo field values
  5. nearSphere_without_index_errors — verifies $nearSphere errors without index (parity with $near)

vic-tsang (changes requested):

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_argument_handling.py
PatersonProjects: Added first four tests, figure #5 and near test would better belong in their respective operators, removed near test form here as such

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_index_behavior.py
PatersonProjects: Moved to errors

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_argument_handling.py
PatersonProjects: Edited to properly reflect test behaviour

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_argument_handling.py
PatersonProjects: Removed inline comments

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_errors.py
```diff
@@ -0,0 +1,525 @@
...
+
+STRICT_CRS = {
+    "type": "name",
+    "properties": {"name": "urn:x-mongodb:crs:strictwinding:EPSG:4326"},
+}
+
+CCW_POLYGON = [[[0, 0], [1, 0], [1, 1], [0, 1], [0, 0]]]
+
+
+INVALID_GEOMETRY_TESTS: list[QueryTestCase] = [
```
fanyangv, line 23: Note: PR #218 ($near) has a test missing_type_defaults_to_point that asserts $geometry without type succeeds (defaults to Point) when used with $near. This test asserts it errors with $geoIntersects. Both pass against MongoDB, so the behavior differs by parent operator. Consider adding a brief comment here noting this is $geoIntersects-specific — $near allows missing type.

documentdb_tests/framework/error_codes.py
```diff
@@ -80,6 +80,7 @@
 MODULO_NON_NUMERIC_ERROR = 16611
 MORE_THAN_ONE_DATE_ERROR = 16612
 CONCAT_TYPE_ERROR = 16702
+CANT_EXTRACT_GEO_KEYS_ERROR = 16755
```
fanyangv: This error code is added but not referenced by any test in this PR. Remove it if unused, or make sure follow-up PR will use it.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_polygon_edge_cases.py
```diff
@@ -0,0 +1,344 @@
...
+            }
+        },
+        doc=[
+            {"_id": 1, "loc": {"type": "Point", "coordinates": [0, 0]}},
+            {"_id": 2, "loc": {"type": "Point", "coordinates": [170, 80]}},
+        ],
+        expected=[{"_id": 2, "loc": {"type": "Point", "coordinates": [170, 80]}}],
+        msg="Polygon larger than hemisphere should be auto-corrected to smaller complement area",
+    ),
+    QueryTestCase(
```
fanyangv: $geoWithin doesn't guarantee document order. This test expects _id: 1 before _id: 2 but that may depend on insertion order. Use ignore_doc_order=True in the assertion to avoid flakiness.

fanyangv (changes requested):

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_errors.py
PatersonProjects: Good catch, I think we're better off removing the case since there will be a test added in geoIntersects looking at #232, it being here leads one to think that's geometry behaviour when it's not.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_polygon_edge_cases.py
PatersonProjects: Added ignore_doc_order

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_errors.py
PatersonProjects: Removed test case

documentdb_tests/framework/error_codes.py
PatersonProjects: Good catch, removed from this PR, to be reintroduced when needed

fanyangv (approved):

eerxuan (changes requested):
Most of the test cases are testing $geometry's container like $geoWithin or $geoIntersects. Those are duplication. This folder should focus on $geometry input validation and GeoJson types. Just use one container in command ($geoIntersects"), don't use others.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_polygon_edge_cases.py
eerxuan, line 13: This is entire file is testing geointersects, e.g. "Should find point inside tall narrow polygon (1 degree wide, 178 degrees tall)". 
Should only keep test cases for $geometry format. Does narrow polygon matter for $geometry? Probably not only the boundary matters.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_coordinate_boundaries.py
```diff
@@ -48,18 +48,6 @@
         expected=[{"_id": 1, "loc": {"type": "Point", "coordinates": [0, -90]}}],
         msg="Should accept latitude = -90",
     ),
-    QueryTestCase(
```
eerxuan: Why removing this?

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_polygon_edge_cases.py
eerxuan: collinear_points_polygon, duplicate_consecutive_vertices,
  many_vertices_polygon, polygon_with_hole, these kind of tests make sense.

PatersonProjects: Thank you for the clarification! Was airing on the side of keep before when in doubt, did a harsher second pass, keeping only those that fully fall within geometry's behaviour.

documentdb_tests/compatibility/tests/core/operator/query/geospatial/specifiers/geometry/test_geometry_coordinate_boundaries.py
PatersonProjects: Got caught incorrectly as other tests were changed from geoWithin to geoIntersects, added back in.

eerxuan (approved):

PR #221, opened 2026-05-21
Add `aggregate` command tests

danielfrankcom:
This change adds tests for the `aggregate` command.

There is some overlap here with #191 as some of the framework changes are needed by both. I've included them here so they should merge cleanly.

PR #225, opened 2026-05-21
Add collation tests

danielfrankcom:
This change adds tests for collation across commands/features. The intent is to cover semantic behavior of collation in general, so these tests cover a number of different features that could potentially be affected by collation.

While working on this change I found some gaps in the collation testing for the `aggregate` command, so will add them in #221. The split is the command tests aim to test the syntax and accepted options for the collation configuration, whereas the dedicated `collation` tests focus on the behavior.

There is some overlap here with #191 as some of the framework changes are needed by both. I've included them here so they should merge cleanly.


documentdb_tests/compatibility/tests/core/collation/command_level/stages/test_stages_accumulators.py
```diff
@@ -0,0 +1,356 @@
+"""Tests for collation effects on $top, $bottom, $topN, $bottomN, $minN, $maxN accumulators."""
```
yshanhu, line 1: To better demonstrate what each test does and ensure test coverage, can we consider a high level folder structure change?
For example:
```
collation/
  ├── utils/
  │   ├── __init__.py
  │   └── collation_view_mismatch.py
  │
  ├── options/
  │   ├── test_collation_locale.py              
  │   ├── ...e.g. strength, caseLevel, caseFirst, numericOrdering, alternate, maxVariable, backwards, normalization
  │   └── test_collation_error_cases.py         # invalid collation docs: missing locale, bad types, unknown fields
  │       # NOTE: each option test uses $match or find to verify behavior
  │
  ├── command_level/
  │   │   # 1-2 tests per operation/stage proving command-level collation works
  │   │   # operations
  │   ├── test_collation_find.py                
  │   ├── test_collation_find_and_modify.py
  │   ├── test_collation_count.py
  │   ...
  │   │   # stages
  │   ├── test_collation_aggregate_match.py     
  │   ├── test_collation_aggregate_sort.py      
  │   ...
  │   └── test_collation_aggregate_graphlookup.py 
  │
  ├── collection_level/
  │   ├── test_collation_collection_default.py  
  │   ├── test_collation_views.py              
  │   ...
  │   # Note: Collation on view should not impact the underlying collection
  │   └── test_collation_view_from_view.py         
  │
  ├── index_level/
  │   ├── test_collation_index.py               # create index with collation, query uses it when collation matches
  │   ...
  │   └── test_collation_index_not_used.py      # query with different collation → COLLSCAN
  │
  └── resolution/
      └── test_collation_resolution.py          # three-way precedence:command-vs-collection-vs-index level collation precedence
          # NOTE: each test uses one operation as example should be good
```

documentdb_tests/compatibility/tests/core/collation/collection_level/test_collection_level_views.py
```diff
@@ -0,0 +1,314 @@
...
+            source_options={"collation": {"locale": "en", "strength": 1}}
+        ),
+        docs=[{"_id": 1, "x": "apple"}, {"_id": 2, "x": "Apple"}],
+        command=lambda ctx: {
+            "aggregate": ctx.collection,
+            "pipeline": [{"$match": {"x": "apple"}}],
+            "cursor": {},
+        },
+        expected=[{"_id": 1, "x": "apple"}],
+        msg="view without collation should not inherit source collection's collation",
```
yshanhu, line 105: Can we add a test case here:
This test verifies that a view doesn't inherit the source collection's collation (collection → view direction). but we're missing the reverse: confirming that a view's collation doesn't affect the underlying collection (view → collection direction).

e.g.
```
  # Setup:
  #   - Collection "orders" created WITHOUT collation (uses binary/"simple")
  #   - View "orders_ci" created on "orders" WITH collation {locale: "en", strength: 1}
  #   - Insert docs: {name: "cafe"}, {name: "Cafe"}, {name: "CAFE"}
  
  # Query the VIEW → uses view's collation (case-insensitive)
  db.orders_ci.find({name: "cafe"})
  
  # Query the COLLECTION → should still use binary, unaffected by view's collation
  db.orders.find({name: "cafe"})
```

documentdb_tests/compatibility/tests/core/collation/command_level/stages/test_stages_accumulators.py
danielfrankcom: Discussed offline, have pushed a change with suggested structure based on this.

documentdb_tests/compatibility/tests/core/collation/index_level/test_index_level_basic.py
```diff
@@ -0,0 +1,521 @@
...
+                {
+                    "key": {"x": 1},
+                    "name": "x_collated",
+                    "collation": {"locale": "en", "strength": 2},
+                }
+            ],
+        },
+        expected={"ok": Eq(1.0)},
+        msg="should create an index with collation specification",
+    ),
```
yshanhu, line 42: Can we add a test case here for creating a second index on the same key with a different collation?
e.g
```
  db.coll_idx_test.drop();
  db.coll_idx_test.createIndex({ name: 1 }, { collation: { locale: "en", strength: 1 }, name: "name_s1" });
  db.coll_idx_test.createIndex({ name: 1 }, { collation: { locale: "en", strength: 2 }, name: "name_s2" });
```

documentdb_tests/compatibility/tests/core/collation/options/test_options_edge_cases.py
```diff
@@ -0,0 +1,168 @@
...
+        expected={"n": 2, "ok": 1.0},
+        msg="count on capped collection should support collation",
+    ),
+]
+
+# Property [Text Index Collation Incompatibility]: a text index cannot be
+# created with a collation other than simple; creating one on a collection
+# with a non-simple default collation requires specifying
+# collation {locale: "simple"} on the index.
+COLLATION_TEXT_INDEX_TESTS: list[CommandTestCase] = [
```
eerxuan: This should be moved to /index_level/

documentdb_tests/compatibility/tests/core/collation/collection_level/test_collection_level_capped.py
```diff
@@ -0,0 +1,168 @@
...
+from documentdb_tests.framework.assertions import assertResult
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+from documentdb_tests.framework.property_checks import Eq
+from documentdb_tests.framework.target_collection import CappedCollection, CustomCollection
+
+# Property [Capped Collection Collation]: a capped collection can be created
+# with a default collation, and collation affects filter matching and sort
+# ordering on capped collections the same as regular collections.
+COLLATION_CAPPED_TESTS: list[CommandTestCase] = [
```
eerxuan, line 19: Move to /collection_level

documentdb_tests/compatibility/tests/core/collation/index_level/test_index_level_basic.py
```diff
@@ -0,0 +1,542 @@
+"""Tests for collation behavior with indexes."""
```
eerxuan, line 1: Missing 2dsphere and 2d index.

documentdb_tests/compatibility/tests/core/collation/collection_level/test_collection_level_views.py
```diff
@@ -0,0 +1,332 @@
+"""Tests for collation constraints on views and cross-view stage behavior."""
+
```
eerxuan, line 2: Timeseries, Clustered not covered.

documentdb_tests/compatibility/tests/core/collation/command_level/operations/test_operations_projection.py
```diff
@@ -0,0 +1,113 @@
+"""Tests for collation effects on positional $ and $elemMatch projection."""
```
eerxuan, line 1: why not put under operations/test_operations_projection_ops.py.

documentdb_tests/compatibility/tests/core/collation/test_collation_dotted_paths.py
```diff
@@ -0,0 +1,280 @@
...
+from documentdb_tests.compatibility.tests.core.collections.commands.utils.command_test_case import (
+    CommandContext,
+    CommandTestCase,
+)
+from documentdb_tests.framework.assertions import assertResult
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+
+# Property [Dotted Path Filter Matching]: collation affects equality and
+# comparison operators on dotted field paths in find and aggregate $match,
```
eerxuan: I think we don't need this file, collation affects on every string comparison. deep field path is implied. One is enough.

documentdb_tests/compatibility/tests/core/collation/test_collation_let_variables.py
```diff
@@ -0,0 +1,166 @@
+"""Tests for collation interaction with command-level let variables."""
```
eerxuan, line 1: $let is expression operator, test_stages_expression.py, and use in update is not needed. We just test one level down: $let + collation. If we test both find and update, we are testing $let + collation + command.

eerxuan (changes requested):

documentdb_tests/compatibility/tests/core/collation/test_collation_let_variables.py
danielfrankcom: I'm not sure I follow this one - `let` in this context is a command argument, it's not the same as the `$let` expression operator. I think it's valuable to do a sanity check here that each of the 3 commands which accept the `let` option support collation properly.

documentdb_tests/compatibility/tests/core/collation/test_collation_dotted_paths.py
danielfrankcom: Moved one representative example to `find`

documentdb_tests/compatibility/tests/core/collation/command_level/stages/test_stages_accumulators.py
danielfrankcom: Resolving for now, please raise another comment if there are further concerns with the latest structure

eerxuan (approved):

eerxuan (approved):

PR #227, opened 2026-05-22
Add capped collection tests

danielfrankcom:
This change adds general tests for capped collections.

The aim here is to test the runtime semantics unique to capped collections like insertion ordering guarantees, natural sort behavior, eviction under size/max constraints, write-position stability across updates and upserts, and delete interactions.

I removed a duplicate smoke test file, and consolidated everything under the `capped_collections` dir. More extensive syntax tests for capped collections with commands like `create` and `convertToCapped` already live elsewhere in test files for those commands. The intent here is to only cover general semantic behavior of capped collections that apply more generally.


documentdb_tests/compatibility/tests/core/collections/capped_collections/test_capped_collections_eviction.py
```diff
@@ -0,0 +1,186 @@
...
+    # Trigger eviction by inserting beyond max.
+    coll.insert_one({"_id": 4, "x": "d"})
+    # The evicted document should no longer be findable via the index.
+    result = execute_command(
+        coll, {"find": coll.name, "filter": {"x": "findme"}, "projection": {"_id": 1}}
+    )
+    assertSuccess(result, [], msg="Evicted documents should not be findable via secondary index")
+
+
+# Property [Size-Based Eviction]: when the total data size exceeds the byte
```
yshanhu, line 64: Can we add tests for eviction after collMod resize? This should be a capped-specific behavior: when limits are reduced via collMod, existing docs survive until the next insert triggers eviction.

e.g
```
#cappedMax
db.createCollection("test", {capped: true, size: 1048576, max: 10})
db.test.insertMany([{_id: 1}, {_id: 2}, {_id: 3}, {_id: 4}, {_id: 5}])
db.runCommand({collMod: "test", cappedMax: 3})
db.test.find().toArray()
Note: We should still have 5 docs here
db.test.insertOne({_id: 6})
db.test.find().toArray()
Note: only 3 docs should be returned
```
Same for cappedSize.

yshanhu (approved):

eerxuan (approved):

PR #229, opened 2026-05-22
Add clustered collection tests

danielfrankcom:
This change adds general tests for clustered collections.

The aim here is to test the runtime semantics unique to clustered collections like `_id` sort behavior, `RecordId` size limits, clustered-specific error codes for special numeric `_id` values, index immutability, and collection-level TTL management via `collMod`.

I renamed the directory from `clustered-collections` to `clustered_collections` to be a valid Python package, and consolidated everything under that directory. More extensive syntax tests for clustered collections with commands like `create` already live elsewhere in test files for those commands. The intent here is to only cover general semantic behavior of clustered collections that apply more generally.


documentdb_tests/compatibility/tests/core/collections/clustered_collections/test_clustered_collections_indexes.py
```diff
@@ -0,0 +1,99 @@
...
+# Property [Explicit _id Index No-Op]: creating a plain {_id: 1} index on a
+# clustered collection is a no-op - the index count does not change.
+@pytest.mark.collection_mgmt
+def test_explicit_id_index_noop(collection):
+    """Test creating {_id: 1} index is a no-op."""
+    name = create_clustered(collection)
+    result = execute_command(
+        collection, {"createIndexes": name, "indexes": [{"key": {"_id": 1}, "name": "test_id"}]}
+    )
+    assertSuccessPartial(result, {"ok": 1.0}, msg="creating {_id: 1} should be a no-op")
```
yshanhu: Minor comment:

Can we consider asserting ``` numIndexesBefore: 0, numIndexesAfter: 0 ```here too to make it a stronger assertion?
This smoke test will show no op + clustered index is not counted in the index count as a clustered-specific behavior.

For example:
```
db.runCommand({drop: "test_count_clustered"})
db.runCommand({create: "test_count_clustered", clusteredIndex: {key: {_id: 1}, unique: true, name: "ci"}})
db.runCommand({createIndexes: "test_count_clustered", indexes: [{key: {x: 1}, name: "x_1"}]})
```

yshanhu (approved):

eerxuan (approved):

PR #233, opened 2026-05-23
Remove materialized views smoke test

danielfrankcom:
It seems materialized views are not a distinct coverage surface, and are effectively just the `merge` and `out` stages. We already have thorough coverage for `out` elsewhere, and `merge` has a placeholder which will be completed later.

I don't think we need this smoke test since it implies a coverage gap that doesn't exist.


eerxuan (approved):
Can you remove the folder and the line item in feature-tree.csv

eerxuan (approved):

eerxuan (approved):

PR #236, opened 2026-05-25
Add timeseries collection tests

danielfrankcom:
This change adds general tests for timeseries collections.

The aim here is to test the runtime semantics unique to timeseries collections. This includes granularity upgrade/downgrade via `collMod`, field immutability (timeField/metaField cannot be changed), default index creation, etc.

A separate file was added under the `createIndexes` command path because timeseries collections have a distinct set of index type restrictions. These are `createIndexes`-specific behaviors that belong alongside the other `createIndexes` tests rather than in the collections directory, consistent with how capped/clustered collection tests don't duplicate command-level validation that already has a better home.

The `TimeseriesCollection` helper in `target_collection.py` was refactored to accept a raw `timeseries_options` dict and optional `create_options` dict, replacing the previous individual fields. This makes it easier to test arbitrary timeseries configurations without needing to add a new field for each option. The existing `collStats` test that used the old interface was updated accordingly.


documentdb_tests/compatibility/tests/core/collections/timeseries/test_timeseries_view_restrictions.py
```diff
@@ -0,0 +1,68 @@
...
+from documentdb_tests.framework.assertions import assertResult
+from documentdb_tests.framework.error_codes import COMMAND_NOT_SUPPORTED_ON_VIEW_ERROR
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+from documentdb_tests.framework.target_collection import TimeseriesCollection
+
+# Property [View-Like Command Rejections]: mapReduce, dataSize, and reIndex
+# are not supported on timeseries collections.
+VIEW_REJECTION_TESTS: list[CommandTestCase] = [
+    CommandTestCase(
```
yshanhu: mapReduce is deprecated, lets remove it

documentdb_tests/compatibility/tests/core/collections/timeseries/test_timeseries_view_restrictions.py
```diff
@@ -0,0 +1,68 @@
...
+        "datasize_rejected",
+        target_collection=TimeseriesCollection(
+            timeseries_options={"timeField": "time", "metaField": "meta"}
+        ),
+        command=lambda ctx: {"dataSize": f"{ctx.database}.{ctx.collection}"},
+        error_code=COMMAND_NOT_SUPPORTED_ON_VIEW_ERROR,
+        msg="Should reject dataSize on timeseries",
+    ),
+    CommandTestCase(
```
yshanhu: reIndex command is removed too, can we remove this test?

documentdb_tests/compatibility/tests/core/indexes/commands/createIndexes/test_createIndexes_timeseries.py
```diff
@@ -0,0 +1,147 @@
...
+    """Test createIndexes rejected index types on timeseries."""
+    result = execute_command(
+        timeseries_coll,
+        {"createIndexes": timeseries_coll.name, "indexes": list(test.indexes)},
+    )
+    assertFailureCode(result, test.error_code, msg=test.msg)
+
+
+# Property [Accepted Index Types]: 2dsphere, partial filter, compound,
+# descending, hidden, and collation indexes are supported on timeseries.
```
yshanhu: sparse and 2d are rejected on measurement fields but allowed on metaField per docs. 
Maybe we can add ```sparse_on_metafield_accepted``` and ```2d_on_metafield_accepted``` cases to show the restriction is field-specific?

```
db.runCommand({createIndexes: "test_ts_pipe", indexes: [{key: {meta: 1}, name: "meta_sparse", sparse: true}]})
db.runCommand({createIndexes: "test_ts_pipe", indexes: [{key: {meta: "2d"}, name: "meta_2d"}]})
```

yshanhu (approved):

eerxuan (approved):

PR #238, opened 2026-05-25
Add views collection tests

danielfrankcom:
This change adds general tests for views.

The aim here is to test the runtime semantics unique to views like source reflection (insert/update/delete visibility), pipeline composition with aggregate queries, view chaining with depth limit enforcement, read operation behavior (find, aggregate, distinct), and write rejection for unsupported operations.

The `ViewCollection` helper in `target_collection.py` was refactored to accept a raw `options` dict, making it easier to test arbitrary view configurations without needing to add a new field for each option.


documentdb_tests/compatibility/tests/core/collections/views/test_views_write_rejection.py
```diff
@@ -0,0 +1,94 @@
...
+        },
+        error_code=COMMAND_NOT_SUPPORTED_ON_VIEW_ERROR,
+        msg="mapReduce on a view should be rejected",
+    ),
+]
+
+
+@pytest.mark.collection_mgmt
+@pytest.mark.parametrize("test", pytest_params(VIEWS_WRITE_REJECTION_TESTS))
+def test_views_write_rejection(database_client, collection, test):
```
yshanhu, line 93: ```createIndexes``` and ```dropIndexes``` are also rejected on views.
Lets consider adding them here:

e.g.
```
db.idxSource.drop()
db.idxSource.insertMany([{_id: 1, x: 10}])
db.createView("idxView", "idxSource", [])

#The following operations should error out
db.runCommand({createIndexes: "idxView", indexes: [{key: {x: 1}, name: "x_1"}]})
db.runCommand({dropIndexes: "idxView", index: "*"})
```

documentdb_tests/compatibility/tests/core/collections/views/test_views_write_rejection.py
```diff
@@ -0,0 +1,94 @@
...
+        docs=[{"_id": 1, "x": 10}],
+        command=lambda ctx: {
+            "findAndModify": ctx.collection,
+            "query": {"_id": 1},
+            "update": {"$set": {"x": 99}},
+        },
+        error_code=COMMAND_NOT_SUPPORTED_ON_VIEW_ERROR,
+        msg="findAndModify on a view should be rejected",
+    ),
+    CommandTestCase(
```
yshanhu, line 66: Can we remove mapReduce case?
mapReduce was removed in MongoDB 8.0. Per test coverage guidelines, deprecated/removed features should not be tested.

documentdb_tests/compatibility/tests/core/collections/views/test_views_chaining.py
```diff
@@ -0,0 +1,90 @@
...
+        result,
+        [{"_id": 2, "x": 20}, {"_id": 3, "x": 30}],
+        msg="view-on-view should compose both pipelines correctly",
+    )
+
+
+# Property [Depth Limit Exceeded]: querying a view chain deeper than 20
+# levels produces the view-depth-limit error.
+@pytest.mark.collection_mgmt
+def test_views_depth_limit_exceeded(database_client, collection):
```
yshanhu: Question: did we check if this exact view depth(i.e. 20) applies across engine types?

danielfrankcom: I think DDB has a slightly different limit from this (22 or something), but it's hard to tell whether this is a compatibility surface worth testing or not. I think the important thing to verify here is that at a certain point this error is returned without allowing unlimited depth, but I don't know how we reconcile the difference for these tests if the limit is intentionally different.

yshanhu: As discussed offline, since the depth limit boundary differs between engines, lets remove this test

yshanhu (approved):

documentdb_tests/compatibility/tests/core/collections/commands/utils/command_test_case.py
```diff
@@ -97,7 +97,7 @@ def prepare(self, db: Database, collection: Collection) -> Collection:
                 target.insert_many(self.docs)
         if self.siblings:
             for sibling in self.siblings:
-                sibling.create(db, collection)
+                sibling.create(db, resolved)
```
eerxuan, line 105: What's the purpose of this change? Can you update docstring of SiblingCollection and CommandTestCase Attributes is not complete.

eerxuan (approved):

documentdb_tests/compatibility/tests/core/collections/commands/utils/command_test_case.py
danielfrankcom: This was fixing a bug from a previous merge which affected the tests in this PR. Will update the docstrings

eerxuan (approved):

PR #241, opened 2026-05-26
Add tailable cursor tests

danielfrankcom:
This change adds tests for general tailable cursor behavior.

The intent here is to cover cross-command behavior and state of the tailable cursor itself, without worrying about command-level input/output validation that is already or will be tested elsewhere.


documentdb_tests/compatibility/tests/core/cursors/tailable_cursors/test_tailable_cursors_awaitdata.py
```diff
@@ -0,0 +1,192 @@
...
+        {"find": capped.name, "tailable": True, "awaitData": True, "batchSize": 100},
+    )
+    cursor_id = result["cursor"]["id"]
+
+    gm_result = execute_command(
+        capped, {"getMore": cursor_id, "collection": capped.name, "maxTimeMS": 200}
+    )
+    assertProperties(
+        gm_result,
+        {"cursor": {"id": Ne(INT64_ZERO), "nextBatch": Eq([])}},
```
yshanhu, line 104: ```test_tailable_cursors_awaitdata_blocks``` and ```test_tailable_cursors_awaitdata_nonblocking``` have identical assertions, so both tests might pass with/without blocking?

To test it properly, can we add a timing assertion to differentiate from the non-blocking test?

documentdb_tests/compatibility/tests/core/cursors/tailable_cursors/test_tailable_cursors_invalidation.py
```diff
@@ -0,0 +1,233 @@
...
+    cursor_id = result["cursor"]["id"]
+
+    # Insert from the same client but different collection handle
+    capped.insert_one({"_id": 1, "src": "conn1"})
+
+    # Use the engine_client to get a second handle to the same collection
+    capped2 = engine_client[capped.database.name][capped.name]
+    capped2.insert_one({"_id": 2, "src": "conn2"})
+
+    # getMore should see both documents
```
yshanhu, line 226: getMore should see both documents but the assertion only checks cursor is open here?

It doesn't verify the 2 inserted documents are actually in nextBatch, can we assert on batch length?

Otherwise, looks good!

documentdb_tests/compatibility/tests/core/cursors/tailable_cursors/test_tailable_cursors_awaitdata.py
danielfrankcom: I did consider testing timing here, but I think it would be too flaky/unreliable. The intent here is only really to test the inputs/outputs, I don't think we care about the specific behavior. I've fixed the naming/wording to make this a bit clearer.

yshanhu (approved):

eerxuan (approved):

PR #253, opened 2026-05-29
Added index type tests for single

vic-tsang:
This PR contains: 
- indexes type tests for single

Note: I realized a lot of tests already exists in createIndexes. It might seem like we are missing some tests but they could be covered already in createIndexes. Rather than having duplicated tests I have omitted them. 

Ref:
- Issue #27


documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_bson_type_validation.py
```diff
@@ -0,0 +1,73 @@
...
+SINGLE_KEY_SORT_ORDER_PARAMS = [
+    BsonTypeTestCase(
+        id="sort_order",
+        msg="single key sort order should reject non-numeric types",
+        keyword="sort_order",
+        valid_types=[BsonType.DOUBLE, BsonType.INT, BsonType.LONG, BsonType.DECIMAL],
+        default_error_code=CANNOT_CREATE_INDEX_ERROR,
+        valid_inputs={
+            BsonType.DOUBLE: 1.0,
+            BsonType.INT: 1,
```
manojnks, line 31: why is INT(1) would fail

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_creation.py
```diff
@@ -0,0 +1,91 @@
...
+def test_single_creation_success(collection, test):
+    """Test single field index creation with valid arguments."""
+    result = execute_command(
+        collection,
+        {"createIndexes": collection.name, "indexes": list(test.indexes)},
+    )
+    assertSuccessPartial(result, index_created_response(), test.msg)
+
+
+def test_single_creation_on_nonexistent_collection(collection):
```
manojnks: this is the same as above

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
```diff
@@ -0,0 +1,192 @@
...
+@pytest.mark.parametrize("test", pytest_params(FIELD_QUERY_TESTS))
+def test_single_field_query(collection, test):
+    """Test single field index queries on various field types."""
+    execute_command(
+        collection,
+        {"createIndexes": collection.name, "indexes": list(test.indexes)},
+    )
+    collection.insert_many(list(test.doc))
+    result = execute_command(collection, {"find": collection.name, "filter": test.filter})
+    assertSuccess(result, test.expected, msg=test.msg)
```
manojnks, line 103: does it assert index usage?

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
```diff
@@ -0,0 +1,192 @@
...
+    collection.insert_many([{"_id": 1, "a": 10}, {"_id": 2, "a": 20}])
+    execute_command(
+        collection,
+        {"createIndexes": collection.name, "indexes": [{"key": {"a": 1}, "name": "a_1"}]},
+    )
+    result = execute_command(
+        collection,
+        {"find": collection.name, "filter": {"a": 10}, "projection": {"_id": 0, "a": 1}},
+    )
+    assertSuccess(result, [{"a": 10}], msg="Should return only indexed field")
```
manojnks: non of the assert actually assert index usage.

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_errors.py
```diff
@@ -0,0 +1,103 @@
...
+        {
+            "createIndexes": collection.name,
+            "indexes": [{"key": {"a": 1}, "name": "a_1", "unique": True}],
+        },
+    )
+    assertFailureCode(
+        result,
+        INDEX_KEY_SPECS_CONFLICT_ERROR,
+        msg="Same key+name with different options should fail",
+    )
```
manojnks: where are partial index , sparse and hidden properties tested for index creation success and failure.

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
```diff
@@ -0,0 +1,192 @@
...
+    collection.insert_many([{"_id": 1, "a": 10, "b": "x"}, {"_id": 2, "a": 20, "b": "y"}])
+    execute_command(
+        collection,
+        {"createIndexes": collection.name, "indexes": [{"key": {"a": 1}, "name": "a_1"}]},
+    )
+    result = execute_command(
+        collection,
+        {"find": collection.name, "filter": {"a": 10}, "projection": {"_id": 0, "a": 1, "b": 1}},
+    )
+    assertSuccess(result, [{"a": 10, "b": "x"}], msg="Should return indexed and non-indexed fields")
```
manojnks: Missing: $in and $ne queries -  these also are used as index filter condition.
Missing: hint usage

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_errors.py
```diff
@@ -0,0 +1,103 @@
...
+    IndexTestCase(
+        id="invalid_empty_field",
+        indexes=({"key": {"": 1}, "name": "empty_1"},),
+        error_code=CANNOT_CREATE_INDEX_ERROR,
+        msg="Empty field name should fail",
+    ),
+]
+
+
+@pytest.mark.parametrize("test", pytest_params(CREATION_ERROR_TESTS))
```
manojnks, line 42: Missing: empty key {}.

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_creation.py
```diff
@@ -0,0 +1,91 @@
...
+    )
+    result = execute_command(
+        collection,
+        {"createIndexes": collection.name, "indexes": [{"key": {"a": -1}, "name": "a_neg1"}]},
+    )
+    assertSuccessPartial(
+        result,
+        {"ok": 1.0, "numIndexesBefore": 2, "numIndexesAfter": 3},
+        msg="Different sort order should create separate index",
+    )
```
manojnks: Missing: dropIndexes after creation.

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_creation.py
```diff
@@ -0,0 +1,91 @@
...
+        id="creation_descending",
+        indexes=({"key": {"a": -1}, "name": "a_neg1"},),
+        msg="Descending order succeeds",
+    ),
+    IndexTestCase(
+        id="creation_dot_notation",
+        indexes=({"key": {"a.b": 1}, "name": "a.b_1"},),
+        msg="Dot notation field succeeds",
+    ),
+]
```
manojnks, line 34: Missing: multiple fields in a single createIndexes call.

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_bson_type_validation.py
vic-tsang: INT 1 in this context is a valid input so it wouldn't failed. It will pass.

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_creation.py
vic-tsang: added multiple single-fiend indexes

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_creation.py
vic-tsang: the non existing is already tested in creation_ascending. Removing this test case

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_creation.py
vic-tsang: I wonder if dropIndexes should be in dropIndexes test ?

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_errors.py
vic-tsang: added

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_errors.py
vic-tsang: these are tested already in functional-tests/documentdb_tests/compatibility/tests/core/indexes/commands/createIndexes.py

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
vic-tsang: when asserting index usage wouldnt that belong to query planner tests ?

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
vic-tsang: same point as above. open for discussion

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
vic-tsang: added

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_creation.py
manojnks: my bad its is mongosh method (https://www.mongodb.com/docs/manual/reference/method/db.collection.createIndexes/) i dont think its applicable here

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_creation.py
manojnks: make sense. will all index type be in it?

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_errors.py
manojnks: ok, sound good

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_bson_type_validation.py
manojnks: ok

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
manojnks: the test has mixed hint and non hint cases. can we add hint to make sure that index is used for the result

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
manojnks: can you talk with yunxuan. but, my opinion is that we should use hint to make sure the created index is okay.

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
eerxuan: We need hint to ask planner to use the created index. So we are testing if the query can use the index. Without hint, planner likely use collection scan and are just testing find query. If we want to test planner, the test would be on larger collection and explain see if planner choose index over collection scan on large collection.

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
eerxuan: Same thing, we need to hint on the index.

eerxuan (commented):
A few cases I'd suggest moving/dropping — most belong in commands/createIndexes/ (or are already there) since
  they aren't single-field-specific. types/single/ should focus on what's unique to a single-field index
  (one-field query semantics, ascending/descending behavior, what BSON values it can index).

  Drop (already covered in createIndexes/):
  - test_single_creation_idempotent → duplicates test_createIndexes_same_key_same_name_noop and
  test_createIndexes_noop_num_indexes_unchanged in naming_and_output.py.
  - invalid_empty_key in test_single_errors.py::CREATION_ERROR_TESTS → covered by empty_key_document in
  test_createIndexes_errors.py:38.
  - CONFLICT_ERROR_TESTS::conflict_same_field_different_name → covered by same_key_different_name in
  test_createIndexes_errors.py:295.
  - CONFLICT_ERROR_TESTS::conflict_same_name_different_field → covered by different_key_same_name in
  test_createIndexes_errors.py:302.

  Move to commands/createIndexes/test_createIndexes_errors.py (gap in conflict matrix, not single-specific):
  - test_single_option_conflict_same_key_name — same key + same name + different options. Genuinely missing from
  createIndexes; please add it there alongside the other same_key_* / different_key_* cases.

  Move to commands/createIndexes/test_createIndexes_naming_and_output.py (or argument_handling.py):
  - test_single_creation_different_sort_creates_two — generic "different directions + different names = both
  coexist" semantic, not single-specific.

  Keep here but fix the assertion:
  - Query tests in test_single_queries.py should pass hint so the assertion verifies the index can answer the
  query, not just that some plan returns the right rows. Without hint, a broken/missing index passes via
  collection scan. (See manojnks's open thread on lines 102/194/208.)
  - test_single_covered_query_* (lines 183-208) — name says "covered" but assertion only checks the projected
  document; a non-covered query passes too. Either drop, or move to a query-planner test that uses explain to
  check winningPlan.stage == PROJECTION_COVERED.

  Keep as-is: the bson_type_validation file, properties.py (NaN/Infinity indexing — genuinely single-field index
  behavior), and the remaining query/sort/descending tests after adding hint.

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
vic-tsang: added hint

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_queries.py
vic-tsang: added

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_creation.py
vic-tsang: yes a variety of indexes type is added. Currently we drop by name only. Do you think we need to drop by specs as well ?

eerxuan: I think we should add them, especially cases like partial index on same indexed fields. 
index 1: {a: 1}, 
index 2: {a: 1}, with partial filter.
Drop {a: 1} what happens,

You can have two indexes with the same key spec if they differ in options (partial filter, sparse, collation, etc.).

eerxuan: not blocking this PR though

eerxuan (approved):

documentdb_tests/compatibility/tests/core/indexes/types/single/test_single_creation.py
vic-tsang: fixed in #580

Issue #43, opened 2026-04-03
Add compatibility test for system: diagnostic commands (19 total)


eerxuan:
system	diagnostic	commands	-	-	buildInfo
system	diagnostic	commands	-	-	collStats
system	diagnostic	commands	-	-	connectionStatus
system	diagnostic	commands	-	-	connPoolStats
system	diagnostic	commands	-	-	dataSize
system	diagnostic	commands	-	-	dbHash
system	diagnostic	commands	-	-	dbStats
system	diagnostic	commands	-	-	explain
system	diagnostic	commands	-	-	getCmdLineOpts
system	diagnostic	commands	-	-	getLog
system	diagnostic	commands	-	-	hostInfo
system	diagnostic	commands	-	-	listCommands
system	diagnostic	commands	-	-	lockInfo
system	diagnostic	commands	-	-	ping
system	diagnostic	commands	-	-	profile
system	diagnostic	commands	-	-	serverStatus
system	diagnostic	commands	-	-	top
system	diagnostic	commands	-	-	validate
system	diagnostic	commands	-	-	whatsmyuri

PR #255, opened 2026-05-29
Add $dbHash tests

PatersonProjects:
This PR adds the tests for the dbHash diagnostic command, as well as a property type check for list inclusion (as opposed to the object inclusion of Contains)

Ref: #43 

Note, next sprint we will be moving away from diagnostic commands to focus on core tests, this was the last one assigned for this sprint.


documentdb_tests/compatibility/tests/system/diagnostic/commands/dbHash/test_dbHash_bson_validation.py
```diff
@@ -0,0 +1,59 @@
+"""Tests for dbHash command BSON type validation."""
```
vic-tsang, line 1: I suggest we rename the file name to bson validation

vic-tsang (changes requested):

documentdb_tests/compatibility/tests/system/diagnostic/commands/dbHash/test_dbHash_bson_validation.py
PatersonProjects: Renamed file

documentdb_tests/framework/property_checks.py
```diff
@@ -164,6 +164,49 @@ def __repr__(self) -> str:
...
+            return f"expected '{path}' to be a dict, got {type(value).__name__}"
+        if self.key not in value:
+            return f"expected '{path}' to contain key '{self.key}'"
+        return None
+
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}({self.key!r})"
+
+
+class Includes(Check):
```
eerxuan: 🟠 Major — Includes vs existing Contains/NotContains API divergence (property_checks.py +43 lines)
  The new Includes class checks if a list contains a literal element, while the existing Contains checks if a list of dicts contains a dict where key=value. The semantics are
  non-overlapping, so a new check is genuinely needed. But the naming reads identically — Contains and Includes are synonyms in English — and the only differentiator is what type of list
  element. A reader hitting Contains(coll.name) vs Includes(coll.name) won't know which one applies without reading both implementations.

  [Suggestion] Either:
  - Rename Includes to something that signals the scalar-element nature, e.g. ContainsElement(x) vs Contains(key, value), or
  - Make Contains polymorphic: Contains(value) for scalar lists, Contains(key=k, value=v) for dict lists.

PatersonProjects: Renamed to ContainsElement

eerxuan (approved):

PR #251, opened 2026-05-28
Add killCursors tests

danielfrankcom:
This change adds tests for the `killCursors` command.

I also fixed a type declaration issue in `documentdb_tests/framework/executor.py` which was causing a type error for one of the tests.

PR #256, opened 2026-05-29
Add getMore tests

danielfrankcom:
This change adds tests for the `getMore` cursor command.

I moved the utils from #251 to a shared location so I could reuse them for this command. Will backport this move to #251.

Closes #348


documentdb_tests/compatibility/tests/core/cursors/commands/getMore/test_getMore_namespace.py
```diff
@@ -0,0 +1,86 @@
...
+        cursor_count=1,
+        find_batch_size=2,
+        command=lambda ctx: {"getMore": ctx.cursors[0], "collection": ctx.collection},
+        expected={"ok": Eq(1.0), "cursor": {"nextBatch": Len(3)}},
+        msg="getMore should succeed when collection matches find cursor namespace",
+    ),
+]
+
+
+@pytest.mark.parametrize("test_case", pytest_params(GETMORE_NAMESPACE_MATCH_FIND_TESTS))
```
ramgokul-14: Does this need to be parameterized at all? Is that the general pattern we follow?

documentdb_tests/compatibility/tests/core/cursors/commands/getMore/test_getMore_max_time_ms.py
```diff
@@ -0,0 +1,387 @@
...
+# Property [maxTimeMS Timeout Retains Cursor]: a getMore that times out with no
+# data keeps the tailable cursor open for subsequent getMore calls.
+def test_getMore_max_time_ms_timeout_retains_cursor(collection):
+    """Test a getMore timeout on a tailable cursor keeps it open for subsequent calls."""
+    capped = CappedCollection().resolve(collection.database, collection)
+    capped.insert_many([{"_id": i} for i in range(5)])
+    cursor_id = open_cursor(capped, {"tailable": True, "awaitData": True, "batchSize": 10})
+    result = execute_command(
+        collection,
+        {"getMore": cursor_id, "collection": capped.name, "maxTimeMS": 0},
```
ramgokul-14, line 361: For my understanding - this produces a timeout? Have we validated that?

ramgokul-14: if it is not actually producing a timeout , should we rename the test or add a comment to explain what exactly we are testing?

ramgokul-14 (changes requested):

documentdb_tests/compatibility/tests/core/cursors/commands/getMore/test_getMore_namespace.py
danielfrankcom: I think this was left over from when there were more tests that I trimmed. Replaced this with an inline method.

documentdb_tests/compatibility/tests/core/cursors/commands/getMore/test_getMore_max_time_ms.py
danielfrankcom: There is a timeout here yes, but the intent is to validate only the contract of the command, and testing the timeout leans a bit too far towards behavioral testing. We've done something similar to this for other commands which have timeouts.

This test is intended to just verify that the cursor is still usable after the timeout occurs. I've tried to clean up the documentation around this so it's a bit clearer.

documentdb_tests/compatibility/tests/core/cursors/commands/getMore/test_getMore_max_time_ms.py
```diff
@@ -0,0 +1,387 @@
...
+            "collection": ctx.collection,
+            "maxTimeMS": 100,
+        },
+        error_code=BAD_VALUE_ERROR,
+        msg="getMore should reject maxTimeMS on a tailable cursor without awaitData",
+    ),
+]
+
+
+@pytest.mark.parametrize("test_case", pytest_params(GETMORE_MAX_TIME_MS_SEMANTIC_TAILABLE_TESTS))
```
ramgokul-14, line 252: This also has just one test in parameter

danielfrankcom: This one I'm less inclined to collapse. It has a slightly different method body from the others so it isn't compatible, but it's following an established pattern of `CursorCommandTestCase` definitions in the file, for those things that can be data-driven. Not all tests fit this format but it's generally easier to parse the test (imo) when they are written this way.

ramgokul-14 (approved):

documentdb_tests/compatibility/tests/core/cursors/commands/getMore/test_getMore_cursor_sources.py
```diff
@@ -0,0 +1,147 @@
...
+                "indexes": [{"key": {f"f{i}": 1}, "name": f"idx_{i}"}],
+            },
+        )
+    list_result = execute_command(
+        collection,
+        {"listIndexes": collection.name, "cursor": {"batchSize": 1}},
+    )
+    cursor_id = list_result["cursor"]["id"]
+    if cursor_id == 0:
+        pytest.skip("listIndexes returned all results in first batch")
```
eerxuan: In what case it would not respect batchSize 1?

danielfrankcom: Good catch, these shouldn't be here, and this is already tested elsewhere.

eerxuan (approved):

PR #259, opened 2026-06-01
Add BSON type testing

danielfrankcom:
This change adds tests for general-purpose BSON type behavior. This differs from operator-specific behavior that exists elsewhere and overlaps with some of the behavior of these types.

As part of this change I moved the existing `core/bson_types/test_bson_types_ordering.py` file into the correct directory, and split it up to better group the tests logically.


documentdb_tests/compatibility/tests/core/bson_types/test_bson_types_ordering.py
```diff
@@ -1,534 +0,0 @@
...
-        doc={},
-        expected=False,
-        msg="Missing nested field not equal to null literal",
-    ),
-]
-
-ALL_INSERT_TESTS = FIELD_PATH_TESTS
-
-
-@pytest.mark.parametrize("test", pytest_params(ALL_INSERT_TESTS))
```
ramgokul-14, line 530: Looks like these tests were not re-added in this PR, is that expected?

danielfrankcom: I'm not sure why these were included in the original test file, but they didn't seem related to BSON types to me so I removed them. Most operator tests already cover field path lookups themselves.

ramgokul-14 (approved):

eerxuan (approved):

Issue #36, opened 2026-04-03
Add compatibility test for update: array operators (8 total)


eerxuan:
core	operator	update	array	-	addToSet
core	operator	update	array	-	pop
core	operator	update	array	-	positional
core	operator	update	array	-	positional-all
core	operator	update	array	-	positional-filtered
core	operator	update	array	-	pull
core	operator	update	array	-	pullAll
core	operator	update	array	-	push

PR #266, opened 2026-06-03
Add positional-all ($[]) tests

PatersonProjects:
This PR adds the tests for the positional-all operator, as well as integration tests with other operators.

Ref: Issue #36


documentdb_tests/compatibility/tests/core/operator/update/utils/__init__.py
```diff
@@ -0,0 +1,3 @@
+from documentdb_tests.compatibility.tests.core.operator.update.utils.update_test_case import (  # noqa: E501, F401
```
vic-tsang, line 3: I think this is not needed. Usually __init__.py is blank

PatersonProjects: Emptied, had to also redirect imports to update_test_case file directly.

documentdb_tests/compatibility/tests/core/operator/update/test_update_array_integration.py
```diff
@@ -0,0 +1,136 @@
...
+from bson import Int64
+
+from documentdb_tests.compatibility.tests.core.operator.update.utils.update_test_case import (
+    UpdateTestCase,
+)
+from documentdb_tests.framework.assertions import assertSuccess
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+
+POSITIONAL_ALL_INTEGRATION_TESTS: list[UpdateTestCase] = [
```
eerxuan, line 21: 🟠 Major — Integration tests verify only command-level success, not output content

  test_update_array_integration.py:17-122. Every test asserts only {"n": 1, "nModified": 1, "ok": 1.0} — never
  verifies the actual array content after the operation. For example inc_all_elements (line 18-25) updates [10,
  20, 30] with $inc: 5 but the assertion doesn't check the result became [15, 25, 35]. The test would pass even
  if $inc silently no-op'd.

  assertSuccess(result, ..., raw_res=True) on the update response confirms only that the command succeeded. To
  verify the operator interaction, the test should also do a follow-up find and assert the document content
  (matching the pattern used in test_positional_all_changed_doc at test_positional_all_core.py:64-77).

  This applies to most of the 12 integration tests. Pick one of:
  - (a) Add a follow-up find + content assertion per case (verbose but explicit), OR
  - (b) Add an expected_doc field to UpdateTestCase and have the integration handler do the find-back
  automatically.

documentdb_tests/compatibility/tests/core/operator/update/array/positional_all/test_positional_all_edge_cases.py
```diff
@@ -0,0 +1,77 @@
...
+        update={"$set": {"arr.$[]": 99}},
+        expected={"_id": 1, "arr": [0, 1, 2]},
+        msg="Query {arr: false} should not match doc with 0 in array (BSON type distinction)",
+    ),
+]
+
+
+LARGE_ARRAY_TESTS: list[UpdateTestCase] = [
+    UpdateTestCase(
+        "large_array_1000_elements",
```
eerxuan: remove large test case.

documentdb_tests/compatibility/tests/core/operator/update/test_update_array_integration.py
PatersonProjects: Refactored, expected is now the expected doc after change, executor function handles the find and assert

documentdb_tests/compatibility/tests/core/operator/update/array/positional_all/test_positional_all_edge_cases.py
PatersonProjects: Removed

eerxuan (approved):

eerxuan (approved):

eerxuan (approved):

PR #560, opened 2026-06-03
Add `commitTransaction` command tests

alinaliBQ:
This change adds tests for the $commitTransaction command operator. 

Add command operator tests for $commitTransaction. Tests database $commitTransaction behavior, output collection, syntax, and expected errors.
- https://github.com/documentdb/functional-tests/issues/411


documentdb_tests/framework/error_codes.py
```diff
@@ -30,6 +30,7 @@
 OPERATION_FAILED_ERROR = 96
 DOCUMENT_VALIDATION_FAILURE_ERROR = 121
 NOT_A_REPLICA_SET_ERROR = 123
+NO_SUCH_TRANSACTION_ERROR = 125
```
Rishabh998: I think this value is wrong. MongoDB's NoSuchTransaction is error code 251, not 125.

https://www.mongodb.com/docs/manual/reference/error-codes/

documentdb_tests/framework/executor.py
```diff
@@ -49,3 +51,68 @@ def execute_admin_command(collection, command: Dict) -> Any:
...
+            collection.insert_many(test_case.docs)
+
+    # 2-4. Transaction lifecycle.
+    client = collection.database.client
+    commit_result = None
+    with client.start_session() as session:
+        session.start_transaction()
+        for op in test_case.ops:
+            op.execute(collection, session)
+        if test_case.commit_command is not None:
```
Rishabh998: When a test supplies its own commit_command, we commit via the raw admin command  but pymongo's session doesn't know we did, so on with block exit it and auto-abort the "still open" transaction.

So for commit_response_ok: insert {_id: 1} → {commitTransaction: 1} returns ok:1 → block exits → pymongo aborts the already-committed txn. The test only checks ok:1, so it'd stay green even if the data got rolled back. Since all the writeConcern / comment / field_types success cases use this path, can you confirm the data actually lands? Reading the doc back after commit (e.g. find_one({"_id": 1})) and asserting it exists would prove it.

documentdb_tests/framework/error_codes.py
alinaliBQ: Yes you are right and I have fixed the naming. The `125` value should actually be named `COMMAND_FAILED_ERROR`, not `NO_SUCH_TRANSACTION_ERROR`.

documentdb_tests/framework/executor.py
alinaliBQ: added check `len(docs) > 0` to verify doc returned persisted

alinaliBQ (commented):
@Rishabh998 Addressed the comments, please have a look

Rishabh998 (approved):

documentdb_tests/compatibility/tests/core/sessions/commands/utils/session_test_case.py
```diff
@@ -0,0 +1,177 @@
...
+                    f"Aborted transaction data persisted — "
+                    f"expected {seed_count} docs (seed only), got {len(docs)}"
+                )
+            else:
+                # Verify that committed data actually persisted (the raw
+                # admin command path bypasses pymongo's session bookkeeping,
+                # so the driver may auto-abort on exit; the abort is a
+                # server no-op after a real commit, but we assert
+                # persistence explicitly).
+                assert len(docs) > 0, (
```
eerxuan: We should not use raw assert here. Every assertion helper have to go through assertFailure or assertSuccess. Because we need the test result categorization in them. 

This also violates the one assertion per test rule.

documentdb_tests/compatibility/tests/core/sessions/commands/utils/session_test_case.py
```diff
@@ -0,0 +1,177 @@
...
+        if not isinstance(command_result, Exception) and test_case.ops:
+            readback = execute_command(
+                collection,
+                {"find": collection.name, "filter": test_case.readback_filter},
+            )
+            assert not isinstance(
+                readback, Exception
+            ), f"Readback after {'abort' if abort else 'commit'} failed: {readback}"
+            cursor = readback.get("cursor", {})
+            docs = cursor.get("firstBatch", [])
```
eerxuan: Where is using readback_filter? Why you verify read back when you want to assert expected_response?

documentdb_tests/compatibility/tests/core/sessions/commands/commitTransaction/test_commitTransaction.py
```diff
@@ -0,0 +1,145 @@
...
+        expected=[{"_id": 1, "x": "after"}],
+        msg="commitTransaction with writeConcern should persist changes",
+    ),
+]
+
+
+@pytest.mark.parametrize("test", pytest_params(COMMIT_PERSISTENCE_TESTS))
+def test_commitTransaction_persistence(collection, test):
+    """Test commitTransaction persists operations."""
+    result = execute_session_command(collection, test)
```
eerxuan: The test run is not readable. Everything obscured in execute_session_command and test case.

eerxuan (commented):
[Suggestion] Restructure the shared session helper — it obscures each test's setup/assert flow

  Thanks for the thorough API-acceptance coverage here — the field_types / comment / writeConcern / *_error files are clean, one-assertion-per-test, and I don't think they need changes.

  My concern is utils/session_test_case.py. execute_session_command bundles the lifecycle, branching, and assertions into one helper, and for only a handful of behavioral tests it carries a lot of complexity:

  - Multiple if/else paths (commit_command set vs. not, expected_response vs. readback, abort vs. commit).
  - Raw assert statements inside the helper (assert len(docs) > 0, …), which both violate TEST_FORMAT.md ("use helpers from framework.assertions, not plain assert") and split each test's real pass/fail logic across two layers — so you can't read a def test_* and see what it verifies.

  Net effect: the actual execution + assertion path is hidden in the helper, and a reader can't tell what any given test is doing.

  Suggested format: inline the lifecycle, one assertion per test

  The transaction lifecycle is only three lines — put them in the test body so setup → act → assertion is visible top to bottom, no helper needed:

  def test_commit_persists_insert(collection):
      """A committed insert is visible outside the transaction afterward."""
      with collection.database.client.start_session() as session:
          session.start_transaction()
          collection.insert_one({"_id": 1, "x": 1}, session=session)
          session.commit_transaction()
      readback = execute_command(collection, {"find": collection.name, "filter": {}})
      assertSuccess(readback, [{"_id": 1, "x": 1}])

  def test_insert_rolled_back_on_abort(collection):
      """An insert is gone after the transaction aborts."""
      with collection.database.client.start_session() as session:
          session.start_transaction()
          collection.insert_one({"_id": 1, "x": 1}, session=session)
          session.abort_transaction()
      readback = execute_command(collection, {"find": collection.name, "filter": {"_id": 1}})
      assertSuccess(readback, [])

  The with is just to close the session; the test itself calls commit_transaction() / abort_transaction(). execute_session_command, the SessionOp enum, and SessionOperation can all go away.

  Use a single test function per spec — no parametrize, and insert is enough

  These are behavioral specs, not type matrices, so one plain def test_* each. We don't need to re-run each spec for insert/update/delete — a simple insert proves the behavior; operation-type coverage belongs in the separate file below.

  Missing behavioral specs

  The PR covers the command API (which inputs are accepted). It doesn't cover full transaction behavior. Listing all specs needed, may exist in this PR or the abortTransaction PR:
  ### Missing behavioral specs (one file per spec, one assertion per test, insert-only)

  - [ ] **Commit persists** — committed insert is visible outside the txn after commit
  - [ ] **Isolation (outside)** — pending insert is NOT visible to a reader outside the txn before commit
  - [ ] **Visibility (in-session)** — pending insert IS visible to a read in the same session before commit
  - [ ] **Rollback on abort** — aborted insert leaves no trace
  - [ ] **Commit is retryable** — re-sending `commitTransaction` after a successful commit returns `ok:1` (committed state is retained for retryability), not an error
  - [ ] **Cannot commit after abort** — commit after abort fails with `251 / NoSuchTransaction`
  - [ ] **Failed op blocks the txn** — after a duplicate-key error (11000) inside the txn, a later op in the same txn fails with `251 / NoSuchTransaction` (must abort to recover)
  - [ ] **Abort recovers the session** — after a failed op + abort, the same session can run a new transaction successfully
  - [ ] **Write conflict** — a second writer to a doc already modified in an open txn fails with `112 / WriteConflict` (first-committer-wins, not blocking)
  - [ ] **Explicit writeConcern on an op inside a txn is rejected** — `72 / InvalidOptions` (the doc: "Setting write concerns for the individual write operations inside a transaction returns an error")
  - [ ] **Cross-collection atomicity** — writing to two different collections in one txn then committing persists both; aborting persists neither (the PR's `commit_multi_operation` only touches one collection)
  - [x] Commit empty transaction — already covered, keep


Separate file for supported/disallowed operations in a transaction
Which operation types are allowed in a transaction is its own concern — put it in a dedicated file rather than folding it into each behavioral test. One test confirming a CRUD op (e.g. find) runs in a txn, and tests that disallowed ops (count, listCollections, listIndexes, explain, createUser) fail.

  ### Disallowed operations inside a transaction
  - [ ] **Write to `config` / `admin` / `local` db** → `263 / OperationNotSupportedInTransaction`
  - [ ] **Write to a capped collection** → `263 / OperationNotSupportedInTransaction`
  - [ ] **`count` command** → rejected (use `$count` aggregation instead)
  - [ ] **`listCollections` / `listIndexes`** → rejected
  - [ ] **`explain`** → rejected (explicitly not allowed in a txn per the doc)
  - [ ] **`createUser`** → rejected

  ### Special Supported operations inside a transaction (separate file)
  - [ ] **`aggregate` `$count`** runs in a txn (the supported counterpart to the disallowed `count` command)

documentdb_tests/compatibility/tests/core/sessions/commands/utils/session_test_case.py
alinaliBQ: I have removed this block of code. 
The readback is being checked in `COMMIT_PERSISTENCE_TESTS`.

documentdb_tests/compatibility/tests/core/sessions/commands/utils/session_test_case.py
alinaliBQ: removed this block of code as it is not required.

documentdb_tests/compatibility/tests/core/sessions/commands/commitTransaction/test_commitTransaction.py
alinaliBQ: `execute_session_command` is removed. Updated to show the command execution

eerxuan (approved):
Please resolve conflict

eerxuan (approved):

PR #561, opened 2026-06-03
Add querySettings aggregation stage tests

### LINKED ISSUE #368: Add compatibility test for $querySettings
Part of #34.

Category: core / operator / stages

Add compatibility test coverage for the `querySettings` feature.

Note: a smoke test exists; comprehensive compatibility test coverage is needed.

danielfrankcom:
This change adds tests for the `querySettings` aggregation stage.

Closes #368


documentdb_tests/compatibility/tests/core/operator/stages/test_stages_position_querySettings.py
```diff
@@ -0,0 +1,249 @@
...
+
+# Property [Subsequent Write Stages]: $out and $merge can follow
+# $querySettings, writing the output stream to a user database without
+# error.
+QUERYSETTINGS_SUBSEQUENT_WRITE_TESTS: list[CommandTestCase] = [
+    CommandTestCase(
+        "subsequent_out_to_user_db",
+        command=lambda ctx: {
+            "aggregate": 1,
+            "pipeline": [{"$querySettings": {}}, {"$out": {"db": ctx.database, "coll": "qs_out"}}],
```
ramgokul-14, line 76: should this (and merge) collection names also be randomized?

ramgokul-14 (commented):
The success case uses replica set but the failure cases do not. Would it make sense to use replica set for the error cases too?

documentdb_tests/compatibility/tests/core/operator/stages/test_stages_position_querySettings.py
danielfrankcom: I don't think it's necessary since every test gets an isolated database which is cleaned up by the pytest fixtures after the test case completes.

ramgokul-14 (approved):

documentdb_tests/compatibility/tests/core/operator/stages/querySettings/test_querySettings_structure.py
```diff
@@ -0,0 +1,214 @@
...
+        admin.command({"removeQuerySettings": query})
+
+
+@pytest.mark.aggregate
+@pytest.mark.replica_set
+@pytest.mark.parametrize("test_case", pytest_params(QUERYSETTINGS_STRUCTURE_TESTS))
+def test_querySettings_structure(collection: Collection, test_case: QuerySettingsStructureTestCase):
+    """Test $querySettings returned document structure."""
+    query_settings = test_case.query_settings(collection)
+    _setup_query_settings(collection, query_settings, test_case.settings_comment)
```
eerxuan: _setup_query_settings failure leaks settings cluster-wide

  test_querySettings_structure.py:201-220 — if setQuerySettings succeeds for query 0 and fails for query 1 inside _setup_query_settings, the exception propagates before the try: block, so the finally block never
  calls removeQuerySettings for query 0. The orphaned setting then survives all subsequent tests in the run. Move the try/finally to wrap the setup call too

documentdb_tests/compatibility/tests/core/operator/stages/querySettings/test_querySettings_type_errors.py
```diff
@@ -0,0 +1,139 @@
...
+        ("binary", Binary(b"x")),
+        ("binary_uuid", Binary(uuid.uuid4().bytes, subtype=4)),
+        ("regex", Regex(".*")),
+        ("code", Code("function(){}")),
+        ("minkey", MinKey()),
+        ("maxkey", MaxKey()),
+    ]
+]
+
+# Property [showDebugQueryShape Coercion Strictness]: a non-boolean value
```
eerxuan, line 87: showDebugQueryShape BSON-type and coercion matrices duplicate each other

  test_querySettings_type_errors.py:285-343. Both lists assert the same rule (non-bool → TYPE_MISMATCH_ERROR) and use overlapping or identical values:
  The "coercion strictness" property is real but observable from any single non-bool case in the BSON list.

eerxuan (commented):
Please add pytestmark = pytest.mark.no_parallel in all 3 files expect for the error file. They can interact cause flaky.

  [Question] Why two test-case classes for one stage?

  The structure tests subclass StageTestCase and execute via db.command(admin); the position tests use CommandTestCase and execute_command(collection, ...). Both run aggregate. Is the split intentional? If
  CommandTestCase (which already supports callable command/expected and target_collection) covers the structure-test needs, sharing one class would simplify. Not blocking.

documentdb_tests/compatibility/tests/core/operator/stages/querySettings/test_querySettings_type_errors.py
danielfrankcom: Changed the typed values to avoid overlap. I think the different potential coercion values are distinct here and should remain, since some may be coerced while others aren't.

documentdb_tests/compatibility/tests/core/operator/stages/querySettings/test_querySettings_structure.py
danielfrankcom: Fixed, similar problem existed in the smoke test.

eerxuan (approved):

PR #563, opened 2026-06-04
Users/afanankhan/behaviourtestsforpush pull all each position operators

afn7081:
Add compatibility tests covering edge cases and behavioral dimensions for
 array update operators:
 
 - $push: field creation, nested paths, non-array errors, value types
   (subdocument, nested array, null), multi, upsert, findAndModify,
   composition with $inc, same-path conflict with $set
 - $pullAll: all-occurrence removal, numeric type matching, subdocument
   matching, nested paths, noop cases, operand/target errors, multi,
   upsert, findAndModify, composition, conflict
 - $each: multiple values, empty array, mixed types, non-array error,
   $addToSet deduplication, subdocuments, upsert, multi, findAndModify,
   $slice composition
 - $position: beginning/middle/end insertion, negative positions,
   beyond-bounds clamping, Int64 typed value, non-integer errors,
   without-$each semantics, upsert, multi, findAndModify, $slice


eerxuan (changes requested):
Can we have sync up, we have overlap in works.

Issue #37, opened 2026-04-03
Add compatibility test for update: bitwise and modifier operators (5 total)


eerxuan:
core	operator	update	bitwise	-	bit
core	operator	update	modifiers	-	each
core	operator	update	modifiers	-	position
core	operator	update	modifiers	-	slice
core	operator	update	modifiers	-	sort

PR #569, opened 2026-06-04
Add $each tests

PatersonProjects:
This PR adds comprehensive tests for the $each operator, as well as some integration tests with the other modifier operators. 

I'll note that currently some tests are duplicated between each used with $push and $addToSet. I have mostly removed these as out of scope but there are a few cases (most notably with the BSON types tests) that they have different error codes or other differences while I think still being within the scope of $each, rather than their own folders. 

Ref: Issue #388 , #37


eerxuan (commented):
🟡 Minor — Coverage gap: $each rejection outside its allowed hosts

  $each is only valid inside $push and $addToSet.  One rejection test per non-host operator would prove $each's scope. The cleanest:

  - {"$pull": {"arr": {"$each": [1]}}} → error
  - {"$pop": {"arr": {"$each": [1]}}} → error
  - {"$pullAll": {"arr": {"$each": [1]}}} → error

  Note: $set doesn't error — it stores the literal {$each: [...]}. Worth a test.

eerxuan (approved):

PR #572, opened 2026-06-05
Add replaceRoot aggregation stage tests

danielfrankcom:
This change adds tests for the `replaceRoot` aggregation stage.

Closes #372


danielfrankcom:
Pushed a few commits with tweaks after testing `replaceWith` which is similar. The main difference is the introduction of a shared constants file which can be used for both sets of stage tests.

ramgokul-14 (commented):
Should we add tests for empty collection / non-existing collection?

ramgokul-14 (approved):

documentdb_tests/compatibility/tests/core/operator/stages/replaceRoot/test_replaceRoot_limits.py
```diff
@@ -0,0 +1,144 @@
+"""Tests for the $replaceRoot aggregation stage."""
```
eerxuan, line 1: [Required] This PR is severely out of date with main — `git diff upstream/main pr572 --stat` shows ~48K deletions because the branch was forked before many PRs landed (`replaceWith`, `$pop`, `$set`, `$unset`, `$rename`, several aggregation stages, and the `docs/REVIEW.md` guide all show as deleted). The actual content additions look reasonable, but a rebase against main is required before this can be evaluated for merge — otherwise it will revert significant prior work.

danielfrankcom: This is overstated - the only conflict was a small one in `error_codes.py` which is common when merging parallel PRs. Fixed in the merge commit.

documentdb_tests/compatibility/tests/core/operator/stages/replaceRoot/test_replaceRoot_limits.py
```diff
@@ -0,0 +1,144 @@
...
+    ),
+]
+
+# Property [Nesting Limit Error]: a constructed replacement-document literal one
+# level deeper than the maximum nesting depth produces an overflow error.
+REPLACEROOT_NESTING_LIMIT_ERROR_TESTS: list[StageTestCase] = [
+    StageTestCase(
+        "nesting_limit_one_over_rejected",
+        docs=[{"_id": 1}],
+        pipeline=[{"$replaceRoot": {"newRoot": _nest(depth=_MAX_NESTING_DEPTH + 1, leaf=1)}}],
```
eerxuan: I think this is global command BSON nests limit, not replaceRoot specific. The actual limit is like 200 under command bson. Here {aggregate: coll, pipeline: [{"$replaceRoot": {"newRoot": 197 more}}]} has 1 array and 2 obj + 197 nested. Same 200 can be reproduced with find/distinct.
Please move the constant to shared place. I think we don't need to test this under every individual feature. Fine to keep it here. It's very rare, won't be a problem for customer.

eerxuan (approved):

PR #574, opened 2026-06-05
Add $position modifier tests

PatersonProjects:
This PR adds the compatibility tests for the $position update modifier operator. It also adds more integration tests, and therefore may conflict with the $each PR currently open: https://github.com/documentdb/functional-tests/pull/569.

Ref: Issue #37 , #389


eerxuan (commented):
🟡 Minor — negative_zero_same_as_zero is a duplicate — Python int -0 evaluates to 0, so this case is identical to position_0_inserts_at_beginning. Either change to double -0.0 or drop.

  🟡 Minor — int32_max_appends/negative_int32_max_inserts_at_beginning overlap with position_exceeds_length_appends. Could keep as boundary-value verification or drop.

eerxuan (approved):

PR #578, opened 2026-06-08
Add replaceWith aggregation stage tests

danielfrankcom:
This change adds tests for the `replaceWith` aggregation stage.

The content of these tests is similar to #572 but the callable structure is a bit different, and there are a few distinct behaviors. I added a shared constants file in `replaceRoot` since several of the definitions are the same between the operators.

Closes #373


ramgokul-14 (approved):

documentdb_tests/compatibility/tests/core/operator/stages/utils/shared_limits.py
```diff
@@ -0,0 +1,32 @@
+"""Limit constants shared between the $replaceRoot and $replaceWith limit tests."""
```
eerxuan, line 1: This creates a sibling-to-sibling import that ties replaceWith tests to the replaceRoot folder structure — a pre-existing stages/utils/ already exists (it hosts stage_test_case.py). Consider moving shared_limits.py
  there.

eerxuan (approved):

PR #559, opened 2026-06-03
Add elemMatch projection tests

danielfrankcom:
This change adds tests for the `elemMatch` projection operator.

I added `OrderedKeys` here since #264 is not merged yet. They are the same change so it should merge cleanly.

Closes #359

PR #584, opened 2026-06-09
Add redact aggregation stage tests

danielfrankcom:
This change adds tests for the `redact` aggregation stage.

There are a few framework changes here which are pulled in from sibling PRs. None of them are unique to this change.

Closes #371


documentdb_tests/framework/property_checks.py
ramgokul-14, line 27: Should we update TEST_FORMAT.md for this?

documentdb_tests/compatibility/tests/core/operator/stages/redact/test_redact_output.py
```diff
@@ -0,0 +1,163 @@
...
+            "pipeline": test_case.pipeline,
+            "cursor": {},
+        },
+    )
+    assertResult(
+        result,
+        expected=test_case.expected,
+        error_code=test_case.error_code,
+        msg=test_case.msg,
+        ignore_doc_order=True,
```
ramgokul-14: is ignore_doc_order=True needed? Specifically since we are doing "REDACT_ORDER_TESTS"
Also is ignore_doc_order=True needed in other places in this PR?

documentdb_tests/framework/property_checks.py
danielfrankcom: Same logic applies as [here](https://github.com/documentdb/functional-tests/pull/559#discussion_r3365430909), I don't think it's necessary.

documentdb_tests/compatibility/tests/core/operator/stages/redact/test_redact_output.py
danielfrankcom: Good spot, it is better to exclude these. Specifically for the order test it doesn't affect it since the `OrderedKeys` check is verifying the key order _within_ a document, whereas `ignore_doc_order` only affects the relative order of documents themselves. That being said it's better to remove `ignore_doc_order` here as it's more restrictive without.

documentdb_tests/compatibility/tests/core/operator/stages/test_stages_position_redact.py
```diff
@@ -0,0 +1,130 @@
...
+            "pipeline": test_case.pipeline,
+            "cursor": {},
+        },
+    )
+    assertResult(
+        result,
+        expected=test_case.expected,
+        error_code=test_case.error_code,
+        msg=test_case.msg,
+        ignore_doc_order=True,
```
ramgokul-14: Is ignore_doc_order needed here? If so, PR looks good to me

ramgokul-14 (approved):

documentdb_tests/compatibility/tests/core/operator/stages/test_stages_position_redact.py
danielfrankcom: Oh yeah missed this one, just searched the main operator directory

ramgokul-14 (approved):

documentdb_tests/compatibility/tests/core/operator/stages/redact/test_redact_descend_structure.py
```diff
@@ -0,0 +1,82 @@
+"""Tests for $redact $$DESCEND structural traversal.
```
eerxuan, line 1: [Required] This PR is severely out of date with main — `git diff upstream/main pr584 --stat` shows ~40K deletions because the branch was forked before many PRs landed (`REVIEW.md`, `AGENTS.md`, `replaceWith` tests, `$pop`, `$set`, `$unset`, etc. all show as deleted). The redact test content looks well-organized, but a rebase against main is required before merge — otherwise it will revert significant prior work.

danielfrankcom: This is overstated - there are no conflicts with the latest `main`. I used the GH tooling to perform a merge from the latest upstream, but it should have no effect on the merge eligibility.

documentdb_tests/compatibility/tests/core/operator/stages/redact/test_redact_expressions.py
```diff
@@ -0,0 +1,322 @@
...
+        "sentinel_first_prune",
+        docs=[{"_id": 1, "a": 1}],
+        pipeline=[{"$redact": {"$first": [["$$PRUNE"]]}}],
+        expected=[],
+        msg="$redact should honor a sentinel extracted from an array by $first",
+    ),
+    StageTestCase(
+        "sentinel_last_prune",
+        docs=[{"_id": 1, "a": 1}],
+        pipeline=[{"$redact": {"$last": [["$$PRUNE"]]}}],
```
eerxuan, line 295: Add test with constructed "$$PRUNE" string, with strconcat operator from "$$", "PRUNE" as literal, also one with the parts as field reference.

documentdb_tests/compatibility/tests/core/operator/stages/redact/test_redact_output.py
```diff
@@ -0,0 +1,162 @@
...
+        ("code", Code("function(){}")),
+        ("minkey", MinKey()),
+        ("maxkey", MaxKey()),
+        ("null", None),
+    ]
+]
+
+# Property [Prune Across Types]: a document is removed under $$PRUNE regardless
+# of the BSON type of a stored field.
+REDACT_PRUNE_TYPE_TESTS: list[StageTestCase] = [
```
eerxuan, line 86: REDACT_PASSTHROUGH_KEEP_TESTS, REDACT_PASSTHROUGH_DESCEND_TESTS, and REDACT_PRUNE_TYPE_TESTS each inline the identical [(tid, val), ...] list of 18 types; test_redact_validation.py:1073 has a fourth near-copy. Per REVIEW.md §6 ("No code duplication across test files / shared test data in utils/"), hoist the pairs into one
  module-level constant (e.g. a redact utils module) and have the three comprehensions iterate it, so a type added or corrected in one place propagates.

danielfrankcom: I agree with this in theory, but this seems like a wider scope problem that shouldn't be tackled here. There are a pretty good number of operator tests that use a similar pattern of type tests. It seems better to keep the duplication here for consistency, and make it easier to replace these if a framework component is introduced later to solve this.

eerxuan (approved):

PR #588, opened 2026-06-09
Add $planCacheListFilters command tests

alinaliBQ:
This change adds tests for the $planCacheListFilters command operator. 

Add command operator tests for $planCacheListFilters. Tests database $planCacheListFilters behavior, syntax, and expected errors.

- https://github.com/documentdb/functional-tests/issues/406

This PR adds `setup` field in `CommandTestCase`.


documentdb_tests/compatibility/tests/core/query-planning/commands/planCacheListFilters/test_planCacheListFilters_lifecycle.py
```diff
@@ -0,0 +1,242 @@
...
+from documentdb_tests.framework.assertions import assertProperties, assertResult
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+from documentdb_tests.framework.property_checks import Eq, Len
+
+# Property [Multiple Filters — Two Shapes]: planCacheListFilters returns both
+# filters when two different query shapes are set.
+LIST_FILTERS_MULTIPLE_TESTS: list[CommandTestCase] = [
+    CommandTestCase(
+        "two_shapes",
```
Rishabh998, line 20: A query shape is the combination of query + sort + projection + collation, but all the multi-filter tests here only change the query. Could we add a case where the query stays the same but the sort (or projection/collation) differs? That should produce two separate filters, and right now nothing checks that.

documentdb_tests/compatibility/tests/core/query-planning/commands/planCacheListFilters/test_planCacheListFilters_response.py
```diff
@@ -0,0 +1,332 @@
...
+        expected={"filters": IsType("array")},
+        msg="planCacheListFilters filters field should be of type array",
+    ),
+]
+
+# Property [Single Filter Entry]: after setting one filter, the response
+# contains one entry with matching query, sort, projection, and indexes.
+LIST_FILTERS_SINGLE_ENTRY_TESTS: list[CommandTestCase] = [
+    CommandTestCase(
+        "single_filter_entry",
```
Rishabh998, line 47: can we add a test of two queries that differ only in their values map to the same shape (e.g. {a: 1} and {a: 999}), one test that sets a filter, sets it again with a different value for the same field, and confirms we still only get 1 filter back.

documentdb_tests/compatibility/tests/core/query-planning/commands/planCacheListFilters/test_planCacheListFilters_lifecycle.py
alinaliBQ: added `same_query_different_sort` to check for case where the query stays the same but the sort differs

documentdb_tests/compatibility/tests/core/query-planning/commands/planCacheListFilters/test_planCacheListFilters_response.py
alinaliBQ: added test `same_shape_different_value` for this case

alinaliBQ (commented):
@Rishabh998 comments are addressed

Rishabh998 (approved):

eerxuan (approved):

eerxuan (approved):

PR #592, opened 2026-06-10
Add currentOp stage tests

danielfrankcom:
This change adds tests for the `currentOp` system stage.

Closes #378


documentdb_tests/compatibility/tests/core/operator/system-stages/currentOp/test_system_stage_currentOp_live_state.py
```diff
@@ -0,0 +1,500 @@
...
+        # instead of relying on a fixed delay.
+        self._wait_for_op(collection, "query")
+        try:
+            yield
+        finally:
+            thread.join(timeout=10)
+
+
+@dataclass(frozen=True)
+class CurrentOpLiveCase(StageTestCase):
```
ramgokul-14, line 155: Should these all be marked no parallel?

documentdb_tests/compatibility/tests/core/operator/system-stages/currentOp/test_system_stage_currentOp_live_state.py
```diff
@@ -0,0 +1,500 @@
...
+class LiveState:
+    """Default. No additional concurrent operation is in flight."""
+
+    @contextmanager
+    def hold(self, collection) -> Iterator[None]:
+        """Establish the concurrent state, run the body, then tear it down."""
+        yield
+
+    @staticmethod
+    def _wait_for_op(collection, op: str, timeout_seconds: float = 10.0) -> None:
```
ramgokul-14, line 32: I think it would be better to fail assertion within this method itself. Otherwise this method returns silently and the test may or may not fail

documentdb_tests/compatibility/tests/core/operator/system-stages/currentOp/test_system_stage_currentOp_live_state.py
danielfrankcom: Good point, I don't think all of them need to be marked but I marked the ones that do.

ramgokul-14 (approved):

documentdb_tests/compatibility/tests/core/operator/system-stages/currentOp/test_system_stage_currentOp_live_state.py
```diff
@@ -0,0 +1,510 @@
+"""Tests for the $currentOp aggregation pipeline stage against live operation state."""
```
eerxuan, line 1: [Required] This PR is severely out of date with main — `git diff upstream/main pr592 --stat` shows ~21K deletions because the branch was forked before many PRs landed (`REVIEW.md`/`AGENTS.md`, `replaceWith`, `$pop`, `$set`, `$unset`, etc. all show as deleted). The `$currentOp` test content itself looks well-organized, but a rebase against main is required before merge — otherwise it will revert significant prior work.

documentdb_tests/framework/error_codes.py
```diff
@@ -45,6 +45,7 @@
 QUERY_FEATURE_NOT_ALLOWED = 224
 MAX_NESTED_SUB_PIPELINE_ERROR = 232
 CONVERSION_FAILURE_ERROR = 241
+OPERATION_NOT_SUPPORTED_IN_TRANSACTION_ERROR = 263
```
eerxuan, line 56: [Minor] `OPERATION_NOT_SUPPORTED_IN_TRANSACTION_ERROR = 263` is also added by PR #605 (listLocalSessions). Whichever PR lands first will own the addition; the other will hit a merge conflict here. Coordinate with #605 so the constant is added once and re-used.

danielfrankcom: We can't really do this until one or the other is merged. This has been a common pattern across merges and is a result of our current parallel review approach. It will require a rebase, yes, but that's trivial to do after one is merged (it doesn't matter which).

documentdb_tests/compatibility/tests/core/operator/system-stages/currentOp/test_system_stage_currentOp_live_state.py
danielfrankcom: This is overstated - there are no conflicts with the latest `main`. I used the GH tooling to perform a merge from the latest upstream, but it should have no effect on the merge eligibility.

documentdb_tests/compatibility/tests/core/operator/system-stages/currentOp/test_system_stage_currentOp_live_state.py
```diff
@@ -0,0 +1,510 @@
...
+                for _ in cursor:
+                    pass
+            except Exception:
+                pass
+
+        thread = threading.Thread(target=run, daemon=True)
+        thread.start()
+        # Wait until the find has issued its first getMore so $currentOp can
+        # observe it, rather than relying on a fixed delay.
+        self._wait_for_op(collection, "getmore")
```
eerxuan: 1. 🔴 test_..._live_state.py (ActiveGetMore ~L114, InFlightLargeCommand ~L153) — _wait_for_op() is called outside the try/finally, so a poll timeout leaks the daemon thread and its live server cursor into later tests.
  The producer thread is started, then self._wait_for_op(...) runs before try: yield ... finally: thread.join(timeout=10). If the op never becomes observable within 10s, _wait_for_op raises AssertionError and thread.join never executes — the daemon thread keeps iterating its $where-sleeping cursor on the shared client. The next test's $currentOp snapshot then sees a stray
  getmore/query, breaking its absence (expected=[]) or field-shape assertions and producing a failure attributed to the wrong test. Fix: move _wait_for_op inside the try so the finally always joins.

eerxuan (approved):

PR #593, opened 2026-06-10
Add `killSessions` command tests

alinaliBQ:
This change adds tests for the killSessions command operator. 

Add command operator tests for killSessions. Tests database killSessions behavior, output collection, syntax, and expected errors.
- https://github.com/documentdb/functional-tests/issues/413


documentdb_tests/compatibility/tests/core/sessions/commands/killSessions/test_killSessions_core.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+@pytest.mark.no_parallel
+def test_killSessions_real_session(collection):
+    """Test killSessions kills a real active session."""
+    client = collection.database.client
+    session = client.start_session()
+    lsid = session.session_id["id"]
+
+    result = execute_command(collection, {"killSessions": [{"id": lsid}]})
```
Rishabh998: This doesn't really test that the session got killed, just that the command returns ok. Can we prove the kill actually did something?

We can't check by listing sessions (docs say a killed session may still show up). But killing a session closes its open cursors — so: open a cursor in the session, kill it, then getMore should fail

documentdb_tests/compatibility/tests/core/sessions/commands/killSessions/test_killSessions_acceptance.py
```diff
@@ -0,0 +1,234 @@
...
+        msg="killSessions should accept readConcern with null level",
+    ),
+]
+
+# Property [Null Array Element Acceptance]: null elements in the array
+# are silently accepted without error.
+KILLSESSIONS_NULL_ELEMENT_TESTS: list[CommandTestCase] = [
+    CommandTestCase(
+        "element_null_accepted",
+        command=lambda ctx: {"killSessions": [None]},
```
Rishabh998, line 160: Should we add a test with [] to killAllsessions?

alinaliBQ: the smoke test happens to use `[]` which kills all sessions:

https://github.com/documentdb/functional-tests/blob/386e8b41caf90c00b2cb69a63b2ff2fb6b54f16d/documentdb_tests/compatibility/tests/core/sessions/commands/killSessions/test_smoke_killSessions.py#L15-L17

so I think we are covered

documentdb_tests/compatibility/tests/core/sessions/commands/killSessions/test_killSessions_core.py
alinaliBQ: ok, I have changed `test_killSessions_real_session` to check for getMore failure after a session is killed

documentdb_tests/compatibility/tests/core/sessions/commands/killSessions/test_killSessions_acceptance.py
```diff
@@ -0,0 +1,236 @@
...
+from documentdb_tests.framework.test_constants import (
+    DECIMAL128_NEGATIVE_ZERO,
+    DECIMAL128_ZERO,
+    DOUBLE_NEGATIVE_ZERO,
+    DOUBLE_ZERO,
+    INT32_MAX,
+    INT64_ZERO,
+)
+
+pytestmark = pytest.mark.no_parallel
```
alinaliBQ, line 26: I am seeing an increase of failed flaky tests in this PR and the failures are not due to sessions being killed. The `killSession` tests are not using valid UUIDs that could match a real session. And test that kills all sessions has already been marked with `no_parallel` so it should not impact rest of code base. 

I think the increased flaky test failures could be due to CI memory space is not enough to handle the `killSessions` commands in parallel. So I added `no_parallel` to all applicable `killSessions` tests to reduce the chances of flaky tests.

alinaliBQ (commented):
@Rishabh998 addressed comments

Rishabh998 (approved):

documentdb_tests/compatibility/tests/core/sessions/commands/killSessions/test_killSessions_core.py
```diff
@@ -0,0 +1,210 @@
...
+        session=session,
+    )
+    cursor_id = find_result["cursor"]["id"]
+
+    # Kill the session.
+    execute_command(collection, {"killSessions": [{"id": lsid}]})
+
+    # getMore on the killed session's cursor should fail with CursorNotFound,
+    # proving the kill closed the cursor.
+    get_more_result = execute_command(
```
eerxuan: Didn't clean up session if failed.

alinaliBQ: Changed to wrap the block using `with client.start_session() as session:`, so `end_session()` is called implicitly by the context manager.

alinaliBQ (commented):
addressed comment

eerxuan (approved):

PR #599, opened 2026-06-11
Add killAllSessions command tests

alinaliBQ:
This change adds tests for the killAllSessions command operator. 

Add command operator tests for killAllSessions. Tests database killAllSessions behavior, output collection, syntax, and expected errors.
- https://github.com/documentdb/functional-tests/issues/412


documentdb_tests/compatibility/tests/core/sessions/commands/killAllSessions/test_killAllSessions_core.py
```diff
@@ -0,0 +1,252 @@
...
+    result = execute_command(collection, {"killAllSessions": []})
+    assertResult(
+        result,
+        expected={"ok": 1.0},
+        msg="killAllSessions should succeed on second call",
+        raw_res=True,
+    )
+
+
+# Property [Kill After Start]: killAllSessions kills an active session.
```
Rishabh998, line 219: Could we add a targeted variant: open two sessions, killAllSessions: [{user, db}] for one of them, then assert the targeted cursor returns CursorNotFound and the other cursor still works?

Lets make sure the filter  {user, db} is honoured

documentdb_tests/compatibility/tests/core/sessions/commands/killAllSessions/test_killAllSessions_core.py
```diff
@@ -0,0 +1,252 @@
...
+def test_killAllSessions_kill_after_start(collection):
+    """Test killAllSessions kills an active session.
```
Rishabh998, line 221: Also double check if these tests do not effect other parallel running test, we donot want to kill session of a parallel running test here

alinaliBQ: The `pytestmark = pytest.mark.no_parallel` in the file applies to all tests that is defined after it, so `test_killAllSessions_kill_after_start` is not run in parallel to other tests.

documentdb_tests/compatibility/tests/core/sessions/commands/killAllSessions/test_killAllSessions_core.py
alinaliBQ: Added test `test_killAllSessions_targeted_filter` for this. I tried to do the same test case u suggested, but since we have no authenticated users (`authenticatedUsers: []`) in the test set up, no `{user, db}` combination can match — only `killAllSessions: []` works.

alinaliBQ (commented):
@Rishabh998 Addressed comments, let me know if you have any questions

Rishabh998 (approved):

eerxuan (approved):

eerxuan (approved):

Issue #39, opened 2026-04-03
Add compatibility test for core: query and write (8 total)


eerxuan:
core	query-and-write	-	-	-	read-concern
core	query-and-write	-	-	-	write-concern
core	query-and-write	commands	-	-	bulkWrite
core	query-and-write	commands	-	-	delete
core	query-and-write	commands	-	-	find
core	query-and-write	commands	-	-	findAndModify
core	query-and-write	commands	-	-	insert
core	query-and-write	commands	-	-	update

PR #604, opened 2026-06-12
Feature: query-and-write delete command

imforster:
#39 
Test cases: 88
Docs: https://www.mongodb.com/docs/v8.2/reference/command/delete/

Adds compatibility test coverage for the query-and-write delete command, validating its full surface area - from basic document deletion and limit semantics to BSON type distinction, null/missing field matching, ordered/unordered error handling, $expr filter support, hint/let/collation option wiring, and argument type rejection - across document types and edge cases.

88 test cases across 7 files following project test guidelines.

Also renames the query-and-write directory tree to use underscores (query_and_write) for valid Python package naming, and renames read-concern and write-concern subdirectories accordingly.

### DIFF
--- a/documentdb_tests/compatibility/tests/core/query_and_write/commands/delete/test_delete_bulk_operations.py
+++ b/documentdb_tests/compatibility/tests/core/query_and_write/commands/delete/test_delete_bulk_operations.py
@@ -113,7 +113,9 @@ def test_delete_ordered_true_stops_on_error(collection):
         result,
         {
             "n": Eq(1),
-            "writeErrors": [Len(1), Eq([{"index": 1, "code": BAD_VALUE_ERROR}])],
+            "writeErrors": Len(1),
+            "writeErrors.0.index": Eq(1),
+            "writeErrors.0.code": Eq(BAD_VALUE_ERROR),
         },
         msg="delete ordered:true should stop at first error with exactly one writeError",
         raw_res=True,
@@ -139,7 +141,9 @@ def test_delete_ordered_false_continues_on_error(collection):
         result,
         {
             "n": Eq(2),
-            "writeErrors": [Len(1), Eq([{"index": 1, "code": BAD_VALUE_ERROR}])],
+            "writeErrors": Len(1),
+            "writeErrors.0.index": Eq(1),
+            "writeErrors.0.code": Eq(BAD_VALUE_ERROR),
         },
         msg="delete ordered:false should continue past errors and report exactly one writeError",
         raw_res=True,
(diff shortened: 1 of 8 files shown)


documentdb_tests/compatibility/tests/core/query_and_write/commands/delete/test_delete_error_cases.py
```diff
@@ -0,0 +1,225 @@
...
+    assertResult(
+        result,
+        expected=test.build_expected(ctx),
+        error_code=test.error_code,
+        msg=test.msg,
+        raw_res=True,
+    )
+
+
+def test_delete_deletes_non_array_object_fails(collection):
```
eerxuan: `assertExceptionType(result, TypeError, ...)` here (and at line 219) checks a client-side PyMongo `TypeError` raised before the command leaves the driver — these test the driver, not the server. Either remove these or move them to a clearly-labeled client-side rejection section with a docstring explaining why they don't go through `assertResult(error_code=...)` like the rest of the file.

documentdb_tests/compatibility/tests/core/query_and_write/commands/delete/test_delete_bulk_operations.py
```diff
@@ -0,0 +1,176 @@
...
+        },
+    )
+    assertSuccessPartial(
+        result,
+        {"ok": 1.0, "n": 2, "writeErrors": [{"index": 1, "code": BAD_VALUE_ERROR}]},
+        msg="delete ordered:false should continue past errors",
+    )
+
+
+def test_delete_ordered_true_error_at_first(collection):
```
eerxuan, line 153: `assertSuccessPartial` extracts only the keys present in `expected`. Asserting `writeErrors: [{"index": 0, "code": ...}]` will pass even if `writeErrors` actually contains multiple entries — so this test won't fail if subsequent statements somehow also produced errors. To actually pin down "stops immediately on first error" (and similarly "continues but reports exactly one error" at line 158), assert the array length explicitly (e.g. via `Len(1)` or a separate length check).

documentdb_tests/compatibility/tests/core/query_and_write/commands/delete/test_delete_bulk_operations.py
```diff
@@ -0,0 +1,176 @@
...
+            "deletes": [
+                {"q": {"status": "D"}, "limit": 0},
+                {"q": {"status": "A"}, "limit": 1},
+            ],
+        },
+        expected={"ok": 1.0, "n": 4},
+        msg="delete should handle mixed limit:0 and limit:1 in bulk",
+    ),
+    CommandTestCase(
+        "duplicate_target_ordered",
```
eerxuan: The case ID is `duplicate_target_ordered` but the command doesn't set `ordered`. Either rename to drop `_ordered` or add `"ordered": True` to make the ID match the behavior under test.

eerxuan (approved):

eerxuan (approved):

PR #605, opened 2026-06-12
Add listLocalSessions stage tests

danielfrankcom:
This change adds tests for the `listLocalSessions` system stage.

Closes #379


documentdb_tests/compatibility/tests/core/operator/system-stages/listLocalSessions/test_listLocalSessions_invocation_errors.py
```diff
@@ -0,0 +1,116 @@
...
+    result = execute_command(target, test.build_command(ctx))
+    assertResult(
+        result, expected=test.build_expected(ctx), error_code=test.error_code, msg=test.msg
+    )
+
+
+# Property [Transaction Invocation Errors]: $listLocalSessions is rejected when run
+# inside a transaction.
+@pytest.mark.aggregate
+def test_listLocalSessions_in_transaction(collection: Collection):
```
ramgokul-14, line 101: does this need replica set?

documentdb_tests/framework/assertions.py
```diff
@@ -379,6 +379,8 @@ def _walk_path(doc: dict, path: str) -> Any:
     if not path:
         return doc
     current: Any = doc
+    if path == "":
+        return current
```
ramgokul-14: Isn't this the same as lines 379-380?

documentdb_tests/framework/error_codes.py
```diff
@@ -45,6 +46,7 @@
 QUERY_FEATURE_NOT_ALLOWED = 224
 MAX_NESTED_SUB_PIPELINE_ERROR = 232
 CONVERSION_FAILURE_ERROR = 241
+OPERATION_NOT_SUPPORTED_IN_TRANSACTION_ERROR = 263
```
ramgokul-14, line 55: is this unused? Should there be a test for this?

documentdb_tests/compatibility/tests/core/operator/stages/test_stages_position_listLocalSessions.py
```diff
@@ -0,0 +1,223 @@
...
+            command runs inside the first session.
+    """
+
+    num_sessions: int = 1
+
+
+# Property [Downstream Stage Composition]: as a first-stage source, $listLocalSessions
+# emits a document stream that downstream stages transform like any other source. Each case
+# runs inside one or more explicit sessions so the local session cache is populated, and
+# asserts output that a no-op stage could not produce.
```
ramgokul-14, line 125: Should this comment be in both places? lines 109 and in 126

documentdb_tests/compatibility/tests/core/operator/system-stages/listLocalSessions/test_listLocalSessions_source_stage.py
```diff
@@ -0,0 +1,62 @@
...
+            "lastUse": IsType("date"),
+        },
+        msg="$listLocalSessions returns documents shaped as {_id: {id, uid}, lastUse}",
+    )
+
+
+# Property [Database Independence]: $listLocalSessions is instance-local, so it succeeds
+# regardless of which database the aggregate command targets.
+@pytest.mark.aggregate
+@pytest.mark.parametrize("database", ["admin", "config"])
```
ramgokul-14, line 90: should we also have a user database in parameter?

documentdb_tests/compatibility/tests/core/operator/stages/test_stages_position_listLocalSessions.py
```diff
@@ -0,0 +1,223 @@
...
+        },
+        expected={"_id": Exists(), "lastUse": IsType("date"), "marker": Eq("present")},
+        msg="a downstream $addFields should add a computed field while preserving originals",
+    ),
+    ListLocalSessionsCompositionTestCase(
+        "followed_by_count",
+        num_sessions=3,
+        command={
+            "aggregate": 1,
+            "pipeline": [{"$listLocalSessions": {"allUsers": True}}, {"$count": "numSessions"}],
```
ramgokul-14, line 162: Should this and any other tests here be marked no_parallel , especially since we are asking for sessions from all users?

documentdb_tests/compatibility/tests/core/operator/system-stages/listLocalSessions/test_listLocalSessions_source_stage.py
```diff
@@ -0,0 +1,62 @@
...
+            collection,
+            {"aggregate": 1, "pipeline": [{"$listLocalSessions": {}}], "cursor": {}},
+            session=session,
+        )
+    assertResult(
+        result,
+        expected={
+            "_id": {
+                "id": BinarySubtype(UUID_SUBTYPE),
+                "uid": ByteLen(32),
```
ramgokul-14: 1. Should we not test both subtype and len for both?
2. (General comment for tests) shouldn't we assert that the id is the particular session id that we want, to avoid validating on some incorrect session?

documentdb_tests/compatibility/tests/core/operator/system-stages/listLocalSessions/test_listLocalSessions_invocation_errors.py
danielfrankcom: Oh yeah, I think I resolved this wrong before. That is the cause of why [this](https://github.com/documentdb/functional-tests/pull/605#discussion_r3415663108) is unused, which should be used again once this is fixed properly.

documentdb_tests/framework/error_codes.py
danielfrankcom: See [this comment](https://github.com/documentdb/functional-tests/pull/605#discussion_r3415646342), should be used again once I push the fix.

documentdb_tests/framework/assertions.py
danielfrankcom: Yes, think this snuck in from a merge conflict, removed.

documentdb_tests/compatibility/tests/core/operator/system-stages/listLocalSessions/test_listLocalSessions_source_stage.py
danielfrankcom: Updated the comment here, I don't think it's necessary. These are just representatives to test an admin and a non-admin DB, it's not meant to be exhaustive given the operator just ignores the DB anyway.

documentdb_tests/compatibility/tests/core/operator/stages/test_stages_position_listLocalSessions.py
danielfrankcom: I don't think it's necessary here since I wrote the assertions defensively so they would tolerate extra sessions beyond the ones created by the test case. We could add `no_parallel` defensively, but I've been trying to avoid it where it's not necessary since the more tests we have like that the slower the overall run will be.

documentdb_tests/compatibility/tests/core/operator/system-stages/listLocalSessions/test_listLocalSessions_source_stage.py
danielfrankcom: There isn't a good way to check the binary subtype for the `uid` since `pymongo` converts it to a Python-native `bytes` object automatically so we no longer get access to the subtype. I added a dedicated test to make sure it's the same across sessions which will hopefully help validate it a bit more.

I also modified the existing test to filter on the session ID and validate it is exactly equal to the expected value as you mentioned.

ramgokul-14 (approved):

documentdb_tests/compatibility/tests/core/operator/system-stages/listLocalSessions/test_listLocalSessions_argument_errors.py
```diff
@@ -0,0 +1,500 @@
+"""Tests for $listLocalSessions invalid-argument errors: types, fields, and mutual exclusion."""
```
eerxuan, line 1: [Required] This PR is out of date with main — `git diff upstream/main pr605 --stat` shows ~9K deletions because the branch was forked before recent PRs landed (killAllSessions, startSession, etc. all show as deleted). A rebase against main is required before merge — otherwise it will revert significant prior work.

documentdb_tests/compatibility/tests/core/operator/system-stages/listLocalSessions/test_listLocalSessions_source_stage.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+
+
+# Property [Database Independence]: $listLocalSessions is instance-local, so it succeeds
+# regardless of which database the aggregate command targets. Cover both an admin and a
+# non-admin database, since a system stage could plausibly be gated to the admin database;
+# the target database is otherwise immaterial because the stage reads instance-local state.
+@pytest.mark.aggregate
+@pytest.mark.parametrize("database", ["admin", "config"])
+def test_listLocalSessions_database_independence(collection: Collection, database: str):
```
eerxuan, line 91: [Suggestion] `test_listLocalSessions_database_independence` parametrizes over `admin` and `config` system databases via `collection.database.client[database]`, bypassing the test-isolation fixture. Two concerns: (1) the test runs against the shared `admin`/`config` databases — under `-n` parallel runs, multiple workers will be hitting the same system namespaces, and (2) the property "instance-local, not gated to admin" is already covered by the other tests in this file which run against the per-test database. Consider dropping this test, or at minimum tagging it `@pytest.mark.no_parallel`.

documentdb_tests/framework/error_codes.py
```diff
@@ -48,6 +49,7 @@
...
```
eerxuan, line 55: [Minor] `OPERATION_NOT_SUPPORTED_IN_TRANSACTION_ERROR = 263` is also added by PR #592 (currentOp). Whichever PR lands first will own the addition; the other will hit a merge conflict here. Coordinate with #592 so the constant is added once and re-used.

danielfrankcom: We can't really do this until one or the other is merged. This has been a common pattern across merges and is a result of our current parallel review approach. It will require a rebase, yes, but that's trivial to do after one is merged (it doesn't matter which).

documentdb_tests/compatibility/tests/core/operator/system-stages/listLocalSessions/test_listLocalSessions_argument_errors.py
danielfrankcom: This is overstated - there are no conflicts with the latest `main`. I used the GH tooling to perform a merge from the latest upstream, but it should have no effect on the merge eligibility.

documentdb_tests/compatibility/tests/core/operator/system-stages/listLocalSessions/test_listLocalSessions_source_stage.py
danielfrankcom: I'd like to keep this one. It's a standalone test for just the database-independence property, where the other tests in the file cover that incidentally while really focusing on output shape and uid semantics. `config` is just standing in as a representative non-admin database and `admin` covers the system-database case.

On the parallel concern, `$listLocalSessions` only reads and writes nothing to these databases, and the test only checks `ok: 1` rather than the sessions it returns, so workers can't step on each other and I don't think `no_parallel` is needed here.

eerxuan (approved):

PR #610, opened 2026-06-16
Add `setQuerySettings` tests

alinaliBQ:
This change adds tests for the setQuerySettings command operator. In order to define the `SettingsTestCase` class, I renamed `query-planning` to `query_planning`. Reason is that hyphens aren't allowed in identifiers, so directories that need to be imported must use underscores instead.

Add command operator tests for setQuerySettings. Tests database setQuerySettings behavior, output collection, syntax, and expected errors.

- https://github.com/documentdb/functional-tests/issues/409

Edit: `replica_set` is needed because query settings are not supported on standalone instances


documentdb_tests/compatibility/tests/core/query_planning/utils/settings_test_case.py
alinaliBQ, line 1: In order to define the `SettingsTestCase` class, I renamed `query-planning` to `query_planning`. Reason is that hyphens aren't allowed in identifiers, so directories that need to be imported must use underscores instead.

documentdb_tests/compatibility/tests/core/query_planning/commands/setQuerySettings/test_setQuerySettings_behavior.py
```diff
@@ -0,0 +1,376 @@
...
+            }
+        ],
+        msg="removeQuerySettings by hash should succeed",
+    ),
+]
+
+
+@pytest.mark.admin
+@pytest.mark.replica_set
+@pytest.mark.parametrize("test", pytest_params(SET_QUERY_SETTINGS_REMOVE_TESTS))
```
karthikvarma22, line 226: why no_parallel in this CR ?

documentdb_tests/compatibility/tests/core/query_planning/commands/setQuerySettings/test_setQuerySettings_behavior.py
```diff
@@ -0,0 +1,376 @@
...
+    ),
+]
+
+
+@pytest.mark.admin
+@pytest.mark.replica_set
+@pytest.mark.parametrize("test", pytest_params(SET_QUERY_SETTINGS_RESPONSE_TESTS))
+def test_setQuerySettings_response(collection, test):
+    """Test setQuerySettings response structure."""
+    ctx = CommandContext.from_collection(collection)
```
karthikvarma22, line 132: Looks like this skeleton is present in most of files. Do we need a shared runner ?

documentdb_tests/compatibility/tests/core/query_planning/commands/setQuerySettings/test_setQuerySettings_behavior.py
alinaliBQ: `no_parallel` is only used in the smoke test. Perhaps you meant `replica_set`? In that case, replica set is required to enable query settings.

documentdb_tests/compatibility/tests/core/query_planning/commands/setQuerySettings/test_setQuerySettings_behavior.py
alinaliBQ: Could you clarify on what you mean by `shared runner`? Currently CI is set up to use a shared mongodb instance. 
According to the style guide, it says the tests should be as standalone as possible and that helper functions that hide the test logic need to be avoided, so we needed to have separate `def test_*` functions.

karthikvarma22: Got it.

documentdb_tests/compatibility/tests/core/query_planning/commands/setQuerySettings/test_setQuerySettings_behavior.py
karthikvarma22: Ack

karthikvarma22 (approved):

eerxuan (commented):
1. Please use @pytest.mark.no_parallel. The test cases can interact each other.
2. Please rebase and use @pytest.mark.requires(...)

eerxuan (approved):

eerxuan (approved):

PR #614, opened 2026-06-16
Add skip_localhost mark, skip shutdown on local

PatersonProjects:
This PR changes the contest and ini to add the needed mark to prevent the shutdown command from running on localhost devices (determined by checking the uri connected to mongo), helping prevent the destructive shutdown failing local test runs.


documentdb_tests/conftest.py
```diff
@@ -341,6 +343,10 @@ def pytest_collection_modifyitems(session, config, items):
     a target they do not apply to (rather than appearing as skips). A target's
     capabilities are determined by its engine and topology, resolved per target
     at runtime (see ``framework.preconditions``).
+
+    Tests marked 'skip_localhost' are skipped when their target connects to a
+    localhost server -- a safety guard for destructive commands (e.g. shutdown)
+    that behave differently from a local client.
```
eerxuan: Can you remind me what's the behavior on local host what is behavior on remove host? 
CI also skips it:
```
skip_localhost:
  Total:   1
  Passed:  0 (0.0%)
  Failed:  0
  Skipped: 1
```
Then when will we run this?

PatersonProjects: Sorry, this came about as a misunderstanding on my part of how Docker was being run in CI, thank you for catching this. The goal is that it is skipped on local runs and passes on remote runs, as if the test ever fails it kills the connection and ruins the whole test run. I have made this earlier change to be a fallback and introduced a new check that allows CI to run the test.

documentdb_tests/compatibility/tests/core/operator/expressions/arithmetic/subtract/test_subtract_basic.py
ramgokul-14, line 1: Were subtract related files unintentionally added?

PatersonProjects: Yes these are unintentional, must have gotten mixed into my local branches as I merged in the preconfigured changes, removing

PatersonProjects: Removed

eerxuan (approved):

PR #624, opened 2026-06-18
Add `applyOps` tests

alinaliBQ:
Add command operator tests for applyOps, including the smoke test. Tests database applyOps behavior, output collection, syntax, and errors. Add the `ReplicationTestCase` class definition. 
Add and use the `replication` capability.

- https://github.com/documentdb/functional-tests/issues/451


documentdb_tests/compatibility/tests/system/replication/commands/applyOps/test_applyOps_multi_ops.py
```diff
@@ -0,0 +1,324 @@
...
+
+# Property [allowAtomic Effectiveness]: allowAtomic: false allows partial
+# commits when a later operation fails.
+APPLYOPS_ATOMIC_EFFECTIVENESS_TESTS: list[ReplicationTestCase] = [
+    ReplicationTestCase(
+        "atomic_false_partial_commit",
+        docs=[{"_id": 1, "x": 1}],
+        command=lambda ctx: {
+            "applyOps": [
+                {"op": "i", "ns": ctx.namespace, "o": {"_id": 2, "x": 2}},
```
karthikvarma22, line 171: I believe in Mongo, duplicate insert doesn't fail, it upserts it, so allowAtomic true or false wouldnt matter. 

Need a case where we have an op that actually fails, like maybe an 'u' instead of insert., and the earlier op still persists

alinaliBQ: Yes you are right. I added a test `doc_check_atomic_false_partial_commit` where `update` (u) on non-existent id fails, and then checks doc from previous op perisists. `atomic_false_partial_commit` checks the error code is returned from invalid `u` command.

karthikvarma22 (approved):

documentdb_tests/compatibility/tests/system/replication/commands/applyOps/test_applyOps_boolean_coercion.py
```diff
@@ -0,0 +1,83 @@
...
+    CommandContext,
+)
+from documentdb_tests.compatibility.tests.system.replication.utils.replication_test_case import (  # noqa: E501
+    ReplicationTestCase,
+)
+from documentdb_tests.framework.assertions import assertSuccessPartial
+from documentdb_tests.framework.executor import execute_admin_command
+from documentdb_tests.framework.parametrize import pytest_params
+
+pytestmark = [pytest.mark.replica_set, pytest.mark.no_parallel]
```
eerxuan: replace pytest.mark.replica_set with @pytest.mark.requires(...)

documentdb_tests/compatibility/tests/system/replication/commands/applyOps/test_applyOps_multi_ops.py
```diff
@@ -0,0 +1,331 @@
...
+    ctx = CommandContext.from_collection(collection)
+    result = execute_command(collection, test.build_command(ctx))
+    assertResult(
+        result,
+        expected=test.build_expected(ctx),
+        msg=test.msg,
+    )
+
+
+def test_applyOps_cross_namespace_doc_check(database_client, collection):
```
eerxuan: test_applyOps_multi_ops.py:1242 — test_applyOps_cross_namespace_doc_check claims to verify inserts in both namespaces but only asserts on ns1.
  The batch inserts into ns1 and ns2, the test creates and writes both collections, but it only queries coll1/result1 and never reads ns2.

eerxuan (commented):
5. 🟡 test_applyOps_core.py:581 & test_applyOps_multi_ops.py:1228 — use_admin is silently ignored by the doc-check functions.
  Every case in APPLYOPS_DOC_CHECK_TESTS / APPLYOPS_MULTI_OPS_DOC_CHECK_TESTS sets use_admin=False, but these test functions call execute_command unconditionally and never read test.use_admin (unlike test_applyOps_core/_entry_validation/_multi_ops, which branch on it). A maintainer who later sets use_admin=True on a doc-check case would get no behavior change and no
  warning. Either honor the flag or drop it from those cases.

documentdb_tests/compatibility/tests/system/replication/commands/applyOps/test_applyOps_boolean_coercion.py
alinaliBQ: rebased and used `@pytest.mark.requires(...)`

documentdb_tests/compatibility/tests/system/replication/commands/applyOps/test_applyOps_multi_ops.py
alinaliBQ: added test `test_applyOps_cross_namespace_doc_check_ns2`, reason is that the test rule specifies one assertion per test function, so we need to create a new test to check ns2.

alinaliBQ (commented):
@eerxuan Addressed all comments. Added `if test.use_admin check` to  test_applyOps_core.py & test_applyOps_multi_ops.py.

documentdb_tests/compatibility/tests/system/replication/commands/applyOps/test_applyOps_multi_ops.py
eerxuan: You could combine two results like {res1, res2} and assert on it. So we have only one test with one assertion. But it's fine.

eerxuan (approved):

PR #625, opened 2026-06-18
Add compactStructuredEncryptionData administration command tests

imforster:
#416 
Test cases: 36
Docs: https://www.mongodb.com/docs/v8.2/reference/command/compactStructuredEncryptionData/

## Summary

Adds comprehensive tests for `compactStructuredEncryptionData` administration command using proper DRY parametrization patterns.

## Changes

- **error_codes.py**: Add `NOT_ENCRYPTED_COLLECTION_ERROR = 6346807` constant
- **field_validation.py**: Uses `bson_type_validator` for exhaustive compactionTokens type rejection (16 BSON types), plus §19 namespace validation and missing field tests
- **core_behavior.py**: Parametrized `CommandTestCase` tests for non-encrypted collection rejection
- **error_cases.py**: Parametrized tests for unrecognized fields and collection variants (views, capped)
- **edge_cases.py**: Parametrized tests for collection name edge cases and compactionTokens content edge cases
- **smoke test**: Fixed to use `assertFailureCode` instead of checking error message content

## What was tested

All 36 tests pass against MongoDB:
- BSON type exhaustiveness for compactionTokens field
- Wire-protocol namespace validation (§19 representative case)
- Non-encrypted collection rejection (core behavior)
- Collection type variants (view, capped)
- Unrecognized field rejection
- Collection name edge cases (system prefix, dots, dollar)
- CompactionTokens document content edge cases


eerxuan (commented):
§16 happy-path coverage — verified reachable on mongo-replset

  Nice, clean DRY structure throughout (good reuse of bson_type_validator, logical file split, and the smoke-test fix to assertFailureCode). One gap worth addressing before merge.

  Every test here asserts a rejection path. Per TEST_COVERAGE.md §16, a collection command also needs its success path — "primary operation succeeds and returns expected response fields." I checked whether that
  path is actually reachable, since on standalone it isn't (creating a QE collection fails with 6346402 "Encrypted collections are not supported on standalone", so the all-rejection strategy is correct for that
  target).

  It is reachable on the mongo-replset target the harness already supports. Tested against mongo:8.2.4:

  // QE collection created server-side via create + encryptedFields (UUID keyId)
  db.runCommand({
    compactStructuredEncryptionData: "enc",
    compactionTokens: { ssn: <bindata> }
  })
  // →
  { stats: { ecoc: { read: 0, deleted: 0 },
             esc:  { read: 0, inserted: 0, updated: 0, deleted: 0 } },
    ok: 1.0 }

  Two things this surfaces:
  1. Success path is testable and returns a documented stats.{esc,ecoc} shape — assertable with assertProperties.
  2. The current compactionTokens "content edge case" tests (null_token_value, empty_string_key, dot_notation_key, nested_document_value) all short-circuit on 6346807 (not-encrypted) on standalone, so they don't
  actually exercise token parsing. On a QE collection, empty tokens against a queryable field instead returns 7294900 "Compact tokens object is missing Compact token for the encrypted path 'ssn'" — that's the
  genuinely-encrypted analog and the path with real signal.

  Suggested additions, gated with @pytest.mark.requires(queryable_encryption=True) (the capability already exists in preconditions.py) so they run on mongo-replset and deselect on standalone:
  - Success: QE collection + valid bindata token → assert ok: 1.0 and the stats.esc/stats.ecoc structure.
  - Missing-token (7294900): QE collection with a queryable field + empty compactionTokens.

  Notes for whoever picks this up:
  - Server-side create with encryptedFields (keyId as UUID Binary subtype 4) is enough — full client-side auto-encryption (pymongo[encryption]) is not needed for the command path.
  - These must be replica-set gated; QE collection creation itself fails on standalone (6346402).
  - I don't see a QE TargetCollection fixture yet — that small enabling piece is the prerequisite. If adding it is out of scope here, a tracked follow-up referencing this comment is fine so the gap is intentional
  rather than invisible.

  Everything else looks good to me.

documentdb_tests/compatibility/tests/system/administration/commands/compactStructuredEncryptionData/test_compactStructuredEncryptionData_edge_cases.py
```diff
@@ -0,0 +1,137 @@
...
+        msg="compactStructuredEncryptionData should reject dollar-prefixed"
+        " collection names with namespace-not-found",
+    ),
+]
+
+# Property [CompactionTokens Content Edge Cases]: compactStructuredEncryptionData
+# handles various compactionTokens document content correctly.
+COMPACTION_TOKENS_CONTENT_TESTS: list[CommandTestCase] = [
+    CommandTestCase(
+        "null_token_value",
```
eerxuan: The four compactionTokens content "edge case" tests (null_token_value, empty_string_key, dot_notation_key, nested_document_value) in edge_cases.py still all resolve to the same NOT_ENCRYPTED_COLLECTION_ERROR on standalone — they short-circuit before token parsing, so they prove the same thing. Now that the QE file exercises real token validation, consider move to the QE-gated file where the content is actually parsed.

alinaliBQ: Done. I have moved `COMPACTION_TOKENS_CONTENT_TESTS` to QE-gated file and converted tests in QE-gated file to use `CommandTestCase`.

alinaliBQ (commented):
@eerxuan I have addressed comment

eerxuan (approved):

PR #629, opened 2026-06-19
Add autoCompact command tests

danielfrankcom:
This change adds tests for the `autoCompact` administration command.

This command is relatively simple, but since it has server-wide effects, these need to run with the `no_parallel` marker. The effect of this command can also extend beyond the runtime of a test case, so I introduced the `ensure_autocompact_idle` utility so test cases can wait for a stable state before they run.

Closes #415


documentdb_tests/compatibility/tests/system/administration/commands/autoCompact/test_autoCompact_operational.py
```diff
@@ -0,0 +1,78 @@
...
+# than silently overriding the running config.
+@pytest.mark.no_parallel
+def test_autoCompact_reconfigure_conflict(collection):
+    """Test autoCompact rejects a differing reconfigure while enabled."""
+    ensure_autocompact_idle(collection)
+    # Establish a running config. This enable reliably succeeds from the idle
+    # state ensure_autocompact_idle guarantees, installing a persistent config
+    # that the differing enable below then conflicts with.
+    execute_admin_command(collection, {"autoCompact": True, "freeSpaceTargetMB": 30})
+    result = execute_admin_command(collection, {"autoCompact": True, "freeSpaceTargetMB": 50})
```
tonglyu, line 74: can we also add a case that set `freeSpaceTargetMB` the same value, will it error out or not?

tonglyu (changes requested):

documentdb_tests/compatibility/tests/system/administration/commands/autoCompact/test_autoCompact_operational.py
danielfrankcom: Good point, added

tonglyu (approved):

eerxuan (approved):

PR #631, opened 2026-06-19
Add explain diagnostic tests

PatersonProjects:
This PR adds the tests for the explain diagnostic command. Explain is an interesting case, as it has both more internal complexity and crosses more features than many of the other diagnostic commands. It also requires a collection as part of its own function call. Because of this, explain tests primarily rely on CommandTestCase rather than DiagnosticTestCase in order to take advantage of runtime collection resolution. 

Ref: Issue #43, #438


documentdb_tests/compatibility/tests/system/diagnostic/commands/explain/test_explain_errors.py
```diff
@@ -0,0 +1,93 @@
...
+    COMMAND_NOT_FOUND_ERROR,
+    ILLEGAL_OPERATION_ERROR,
+    NO_QUERY_EXECUTION_PLANS_ERROR,
+)
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+
+pytestmark = pytest.mark.admin
+
+EXPLAIN_ERROR_TESTS: list[CommandTestCase] = [
```
fanyangv, line 26: The docs state: "You cannot run explain in `executionStats` or `allPlansExecution` mode for a pipeline containing `$out`." Consider adding a test: explain with `{aggregate: ..., pipeline: [{$out: "target"}]}` and verbosity: `"executionStats"` should error.

documentdb_tests/compatibility/tests/system/diagnostic/commands/explain/test_explain_verbosity_and_response.py
```diff
@@ -0,0 +1,209 @@
...
+)
+from documentdb_tests.framework.assertions import assertResult
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+from documentdb_tests.framework.property_checks import Exists, IsType, NotExists
+
+pytestmark = pytest.mark.admin
+
+
+RESPONSE_FIELD_TESTS: list[CommandTestCase] = [
```
fanyangv, line 23: The docs state the default verbosity (when omitted) is `allPlansExecution`. Consider adding a test that omits verbosity entirely and verifies both `queryPlanner` and `executionStats` are present — confirming the default mode matches `allPlansExecution`.

fanyangv (changes requested):

documentdb_tests/compatibility/tests/system/diagnostic/commands/explain/test_explain_verbosity_and_response.py
PatersonProjects: Added!

documentdb_tests/compatibility/tests/system/diagnostic/commands/explain/test_explain_errors.py
PatersonProjects: Added!

PatersonProjects: Test is failing, thought at first might be a compatibility gap, but it looks like the actual behaviour is to run the command plan up to the $out, rather than error. Looking closely at the docs it never explicitly states it will error, just that you cannot run the mode containing out, so I think it is correct behaviour to silently drop the $out, just misleading docs.

PatersonProjects: Given the above, I have dropped the test, but am leaving this unresolved for your consideration, as changing the test to match the observed behaviour is also an option.

fanyangv: I would suggest to have this test with observed behavior and add comments

fanyangv (changes requested):

documentdb_tests/compatibility/tests/system/diagnostic/commands/explain/test_explain_errors.py
PatersonProjects: Added positive version of test to response file, matching observed behaviour

fanyangv (approved):

eerxuan (commented):
Please also verify queryPlanner.planCacheShapeHash and queryShapeHash exist. 
We need a hash equivalence test for both planCacheShapeHash and queryShapeHash. 
https://github.com/documentdb/functional-tests/issues/438#issuecomment-4697132188

documentdb_tests/compatibility/tests/system/diagnostic/commands/explain/test_explain_queryhash_shape.py
```diff
@@ -0,0 +1,577 @@
...
+    result = execute_command(collection, {"explain": cmd, "verbosity": "queryPlanner"})
+    if isinstance(result, Exception):
+        raise RuntimeError(f"explain command failed: {result}")
+    query_planner = result.get("queryPlanner", {})
+    if "planCacheShapeHash" not in query_planner:
+        raise RuntimeError(f"explain did not return queryPlanner.planCacheShapeHash; got: {result}")
+    return str(query_planner["planCacheShapeHash"])
+
+
+def get_query_shape_hash(
```
eerxuan, line 46: Can you add a few tests to show difference between queryShapeHash and planCacheShapeHash. The planCacheShapeHash only consider the filter, sort, projection, and a few aggregation pipeline. planCacheShapeHash supports all stages.

PatersonProjects: Added tests for various aggregation stages to showcases different cases between queryShapeHash and planCacheShapeHash

eerxuan (approved):

PR #633, opened 2026-06-19
Add getDefaultRWConcern command tests

danielfrankcom:
This change adds tests for the `getDefaultRWConcern` administration command.

Closes #423


documentdb_tests/compatibility/tests/system/administration/commands/getDefaultRWConcern/test_getDefaultRWConcern_response.py
```diff
@@ -0,0 +1,138 @@
...
+        expected=test.expected,
+        error_code=test.error_code,
+        msg=test.msg,
+        raw_res=True,
+    )
+
+
+# Property [Topology Errors]: on a standalone node the command is unsupported
+# and is rejected.
+def test_getDefaultRWConcern_topology_error(collection):
```
tonglyu, line 68: do we have standalone mode marker? should use marker if it cannot be run with replica mode

tonglyu (changes requested):

documentdb_tests/compatibility/tests/system/administration/commands/getDefaultRWConcern/test_getDefaultRWConcern_response.py
danielfrankcom: I considered adding one here but there are other in-flight changes related to this right now, and it will be better to wait. Currently the tests only run against standalone, so it is not an issue to merge this as-is.

Depending on merge order (either this merges first or other PR(s) merge first) one will have to rebase and fix this issue since the test will start failing, so it is self-correcting. We don't have the infrastructure in the framework in the context of this PR to correct this right now, so I don't think we need to block on this.

Related issue: #568
Related PRs: #617, #560

tonglyu (approved):

eerxuan (approved):

PR #637, opened 2026-06-23
Add admin command tests for killOp

vic-tsang:
This PR contains:
- admin command tests for killOp

Ref:
- Issue #425


documentdb_tests/compatibility/tests/system/administration/commands/killOp/test_killOp_core_behavior.py
```diff
@@ -0,0 +1,90 @@
+"""Tests for killOp command core behavior.
+
```
SaiPradyumn, line 2: The most significant gap: no test actually verifies that killOp terminates a running operation. Every
test in this file uses NON_RUNNING_OP_ID and only asserts that killOp returns ok: 1 — which it does regardless of whether any op exists. This tests the response shape, not the core behavior.
Consider adding a test that:
1. Spawns a long-running operation (e.g., $where sleep, or a find with maxTimeMS on a large collection)
2. Captures its opId from currentOp
3. Calls killOp with that opId
4. Verifies the operation was actually interrupted (client receives an error or currentOp no longer lists it)

eerxuan: It's fine to not verify as the expected behavior is straight forward, we don't expect engine to be different. The test will be multi-step and have racing conditions, better to keep tests simple.

vic-tsang: I have confirmed with Yunxuan that we won't be doing multiple step verification.

SaiPradyumn (approved):

SaiPradyumn (approved):

eerxuan (approved):

PR #638, opened 2026-06-23
Add searchMeta stage tests

danielfrankcom:
This change adds tests for the `$searchMeta` aggregation stage.

This is the first operator which uses the external `mongot` topology to provide a required search index. I defined a number of file-scoped pytest fixtures to define the necessary indexed collections for these tests, as an index of this type is far more expensive to set up than the ones we use elsewhere (on the order of 30-60 secs).  Using a module scoped fixture allows us to amortize this cost across the test cases rather than taking the hit each time.

~~**Note:** The CI runs for this change will fail until #617 is merged, as it is heavily dependent. If you want to see the change in its correct (working) context, see the [pr/searchMeta](https://github.com/danielfrankcom/ddb-tests-contrib/tree/pr/searchMeta) branch for now.~~ This is resolved now.

Closes #375


documentdb_tests/compatibility/tests/core/operator/stages/searchMeta/test_smoke_searchMeta.py
tonglyu, line 1: Per the MongoDB docs, $searchMeta accepts concurrent, returnScope, and returnStoredSource fields (same as $search). None of these have any test coverage here.

documentdb_tests/compatibility/tests/core/operator/stages/searchMeta/test_searchMeta_facets.py
tonglyu, line 1: The date facet success tests validate boundary bucketing, but don't cover the date facet default overflow bucket. MongoDB docs show date facets support the same default field as number facets

documentdb_tests/compatibility/tests/core/operator/stages/searchMeta/test_searchMeta_spec_facet_boundary_errors.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+@pytest.fixture(scope="module")
+def search_collection(engine_client, worker_id) -> Iterator[Collection]:
+    """Module-scoped metadata search collection (default + alt_idx indexes)."""
+    with open_search_collection(engine_client, worker_id, f"{__name__}::search_collection") as coll:
+        yield coll
+
+
+# Property [Facet Number Boundaries Validation]: number-facet boundaries must be
```
tonglyu, line 30: can we add
1/ duplicate adjacent boundaries like [0, 0, 25] which violates "distinct ascending"
2/ an array exceeding the documented 1000 element limit.

tonglyu (changes requested):

documentdb_tests/compatibility/tests/core/operator/stages/searchMeta/test_searchMeta_metadata.py
```diff
@@ -0,0 +1,366 @@
+"""Tests for $searchMeta metadata and count result semantics."""
```
tonglyu, line 1: This seems also incomplete list for operators. Can we test against all operators even it's not supported?
Following are missing:  autocomplete, in, regex, geoWithin, geoShape, queryString, moreLikeThis, term, span

documentdb_tests/compatibility/tests/core/operator/stages/searchMeta/test_searchMeta_options.py
```diff
@@ -0,0 +1,198 @@
+"""Tests for $searchMeta stage options (concurrent, returnScope)."""
```
tonglyu: can we also add success path case for returnScope? currently it's most validation the field itself but not functionality

documentdb_tests/compatibility/tests/core/operator/stages/searchMeta/test_searchMeta_metadata.py
danielfrankcom: Yes, had to add some new fixtures to test some of these since they have specific requirements, but there is a new file which covers all known operators now.

documentdb_tests/compatibility/tests/core/operator/stages/searchMeta/test_searchMeta_options.py
danielfrankcom: Yes, moved related tests to a separate file and added these.

tonglyu (approved):

documentdb_tests/compatibility/tests/core/operator/stages/searchMeta/test_searchMeta_return_scope.py
```diff
@@ -0,0 +1,266 @@
...
+from documentdb_tests.framework.error_codes import UNKNOWN_ERROR
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+from documentdb_tests.framework.test_constants import DECIMAL128_ONE_AND_HALF
+
+pytestmark = pytest.mark.requires(search=True)
+
+
+@pytest.fixture(scope="module")
+def search_collection(engine_client, worker_id) -> Iterator[Collection]:
```
eerxuan: 1. searchMeta/utils/ — the plain search_collection module fixture is copy-pasted verbatim across ~8 test files. 🟡 Minor (reuse)

  Roughly eight files each define the identical:
  @pytest.fixture(scope="module")
  def search_collection(engine_client, worker_id) -> Iterator[Collection]:
      with open_search_collection(engine_client, worker_id, f"{__name__}::search_collection") as coll:
          yield coll
  (The other ~19 module fixtures — empty_search_collection, no_index_collection, date_facet_collection, geo_collection, autocomplete_collection, etc. — are genuinely distinct recipes and are not the target here.) The repeated default one could live once in searchMeta/utils/ (or a conftest.py) and be imported. Note the honest
  caveat: module scope means it still instantiates per module at runtime, so this is a maintainability/DRY win (one place to change the default index recipe), not a performance one. Worth doing given the count.

danielfrankcom: Thanks, initially duplicated these because `conftest.py` files were rejected in the test packages, but later decided to change that for `search` and forgot to fix here.

eerxuan (approved):

eerxuan (approved):

PR #639, opened 2026-06-23
Add `validate` tests

alinaliBQ:
Add command operator tests for validate. Tests database validate behavior, output collection, syntax, and errors. 
Adds `setup` field in `DiagnosticTestCase`.

- https://github.com/documentdb/functional-tests/issues/448


documentdb_tests/compatibility/tests/system/diagnostic/commands/validate/test_validate_bool_param_coercion.py
```diff
@@ -0,0 +1,166 @@
...
+from documentdb_tests.compatibility.tests.system.diagnostic.utils.diagnostic_test_case import (
+    DiagnosticTestCase,
+)
+from documentdb_tests.framework.assertions import assertProperties
+from documentdb_tests.framework.executor import execute_command
+from documentdb_tests.framework.parametrize import pytest_params
+from documentdb_tests.framework.property_checks import Eq
+
+# Property [Type Coercion]: validate accepts all BSON types for the full parameter via coercion.
+ACCEPTED_TYPE_TESTS: list[DiagnosticTestCase] = [
```
eerxuan, line 24: 🟠 [Major] The per-parameter coercion files are enormous and partly redundant against §20's canonical matrix. Five files (full, metadata, checkBSONConformance, repair, fixMultikey) each test the same ~21-type coercion matrix (~105 near-identical cases). Per TEST_COVERAGE.md §20, boolean-param coercion should mirror the canonical matrix (bypassDocumentValidation), but it
  doesn't require re-testing all 21 BSON types five times for one command — the coercion logic is shared across these boolean params. Consider collapsing to one representative coercion matrix plus per-param confirmation that each is wired to it (a few cases each). As written this is a lot of duplicated surface for one behavior; TEST_FORMAT.md "don't ask for exhaustive
  BSON matrix on simple shared params" cuts toward trimming.

documentdb_tests/compatibility/tests/system/diagnostic/commands/validate/test_validate_response_structure.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+RESPONSE_DETAIL_TESTS = RESPONSE_VALUE_TESTS + DETAILED_STRUCTURE_TESTS
+
+
+@pytest.mark.parametrize("test", pytest_params(RESPONSE_DETAIL_TESTS))
+def test_validate_response_details(collection, test):
+    """Test validate response values, sub-structure, and option-specific shapes."""
+    for cmd in test.setup:
+        execute_command(collection, {**cmd, next(iter(cmd)): collection.name})
```
eerxuan: [Question] setup runner uses next(iter(cmd)) to rewrite the collection name. The pattern {**cmd, next(iter(cmd)): collection.name} assumes the command name is the first key in each setup dict. That holds for the literal dicts here ({"insert": "", ...}), but it's a slightly fragile idiom — if a dict were ever written with another key first, it'd rewrite the wrong
  field silently. A small explicit helper or a command_name field would be more robust. Minor.

documentdb_tests/compatibility/tests/system/diagnostic/commands/validate/test_validate_bool_param_coercion.py
alinaliBQ: removed the exhaustive tests

documentdb_tests/compatibility/tests/system/diagnostic/commands/validate/test_validate_response_structure.py
alinaliBQ: Switched to use helper `bind_collection`

alinaliBQ (commented):
Addressed comments. also added `validate_repair` capability to skip on `validate` tests that only work on standalone instances

eerxuan (approved):

PR #642, opened 2026-06-24
Add admin command tests for logRotate

vic-tsang:
This PR contains:
- admin command tests for logRotate
-  execute_admin_with_retry_command: added a retry mechanism for execute_admin_command. Without the retry mechanism, the first test would be pass but subsequent tests would fail. 
 
Background: 
- When logRotate runs, it renames the active log file to an archive name built from the
  current timestamp with one-second resolution, e.g. my_server_log.2026-06-24T17-18-27. The rename target is fully determined by the wall-clock second. 
- If two rename happens within the same second, they would both have the same name and one of the rename will fail. 

Discussion Points: 
- Audit log is skipped as it needs a special flag for the server. Please let me know if it worthwhile to enable for one test case. 

Ref:
- Issue #426


documentdb_tests/compatibility/tests/system/administration/commands/logRotate/test_logRotate_bson_type_validation.py
```diff
@@ -0,0 +1,92 @@
...
+    """Test comment field accepts all BSON types."""
+    result = execute_admin_with_retry_command(
+        collection, {"logRotate": 1, "comment": sample_value}, retry_code=FILE_RENAME_FAILED_ERROR
+    )
+    assertSuccessPartial(
+        result,
+        {"ok": 1.0},
+        msg=f"comment should accept {bson_type.value}",
+    )
+
```
SaiPradyumn, line 89: Other admin command tests (killOp, getParameter) assert "comment": NotExists() in the response. Here only {"ok": 1.0} is checked. Add NotExists() to verify the comment is consumed and not echoed back

documentdb_tests/compatibility/tests/system/administration/commands/logRotate/test_logRotate_errors.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_logRotate_unrecognized_field(collection):
+    """Test that logRotate rejects unrecognized top-level fields."""
+    result = execute_admin_command(collection, {"logRotate": 1, "unknownField": 1})
+    assertFailureCode(
+        result, UNRECOGNIZED_COMMAND_FIELD_ERROR, msg="Should reject unrecognized fields"
+    )
+
+
```
SaiPradyumn, line 62: Missing: does an unrecognized field still fail when a valid comment is also present?

documentdb_tests/compatibility/tests/system/administration/commands/logRotate/test_logRotate_bson_type_validation.py
vic-tsang: fixed

documentdb_tests/compatibility/tests/system/administration/commands/logRotate/test_logRotate_errors.py
vic-tsang: added

eerxuan (approved):

PR #645, opened 2026-06-25
Add search stage tests

danielfrankcom:
This change adds tests for the `search` aggregation stage.

The `mongot` sidecar change here overlaps with #638. Both contain the same commit that should merge cleanly regardless of which is merged first.

I defined a number of pytest fixtures to define the necessary indexed collections for these tests, as an index of this type is far more expensive to set up than the ones we use elsewhere (on the order of 30-60 secs). Using a module/package scoped fixture allows us to amortize this cost across the test cases rather than taking the hit each time.

Closes #374


documentdb_tests/compatibility/tests/core/operator/stages/search/test_search_count.py
```diff
@@ -0,0 +1,239 @@
...
+                    "$search": {
+                        "text": {"query": "quick", "path": "title"},
+                        "count": {"threshold": val},
+                    }
+                },
+            ],
+            error_code=UNKNOWN_ERROR,
+            msg=f"$search should reject a {tid} count.threshold as a non-integer",
+        )
+        for tid, val in [
```
tonglyu, line 266: can we also add null/undefined, INF, -INF to see if they are valid?

documentdb_tests/compatibility/tests/core/operator/stages/search/test_search_equals.py
```diff
@@ -0,0 +1,366 @@
...
+# Property [Equals Value Type Rejection]: equals.value rejects any type outside
+# the supported set (bool, objectId, number, string, date, uuid, null).
+SEARCH_EQUALS_VALUE_TYPE_ERROR_TESTS: list[StageTestCase] = [
+    StageTestCase(
+        f"equals_value_type_{tid}",
+        pipeline=[{"$search": {"equals": {"path": "num", "value": val}}}],
+        error_code=UNKNOWN_ERROR,
+        msg=f"$search equals should reject a {tid} value as an unsupported type",
+    )
+    for tid, val in [
```
tonglyu, line 355: do we want to test more types here?

documentdb_tests/compatibility/tests/core/operator/stages/search/test_search_in.py
```diff
@@ -0,0 +1,427 @@
+"""Tests for the $search in operator."""
```
tonglyu, line 1: can we test required-field validation. Per the MongoDB docs, both `path` and `value` are required.

Also, a general comment for this PR: score and doesNotAffect are option field for $in operator. Do we expect test them as well?

These two attributes also missed in operator like near, equals, range, exits, etc

documentdb_tests/compatibility/tests/core/operator/stages/search/test_search_near.py
```diff
@@ -0,0 +1,393 @@
+"""Tests for the $search near operator."""
```
tonglyu, line 1: same here, test `path`,
`origin`, and `pivot` are all required.

tonglyu (changes requested):

documentdb_tests/compatibility/tests/core/operator/stages/search/test_search_count.py
danielfrankcom: Not sure if you were referring to the "undefined" BSON type, but that one is deprecated so have excluded it (as consistent with other tests). Added the others though.

documentdb_tests/compatibility/tests/core/operator/stages/search/test_search_equals.py
danielfrankcom: This is actually all of the types that exhibit this behavior, the others are accepted

documentdb_tests/compatibility/tests/core/operator/stages/search/test_search_in.py
danielfrankcom: I don't think we need the complete testing for score everywhere since it's accepted on ~15 operators, but I added a representative test for each (including `$in`) to verify it's compatible with each operator. The shared `test_search_score.py` file should be sufficient for covering the deeper functionality.

`doesNotAffect` is less widely supported (3 operators only), but I added representative tests for those too.

documentdb_tests/compatibility/tests/core/operator/stages/search/test_search_text_errors.py
```diff
@@ -0,0 +1,398 @@
...
+    StageTestCase(
+        "text_path_object_no_value",
+        pipeline=[{"$search": {"text": {"query": "quick", "path": {}}}}],
+        error_code=UNKNOWN_ERROR,
+        msg="$search should reject a text.path document with no value field",
+    ),
+    StageTestCase(
+        "text_path_object_absent_multi",
+        pipeline=[
+            {"$search": {"text": {"query": "quick", "path": {"value": "title", "multi": "nope"}}}},
```
tonglyu, line 146: It seems like {value, multi} multi-analyzer path form has only a negative case, can we add success case for it?

tonglyu (approved):

eerxuan (commented):
Please check the test failure, failed on retry.

eerxuan (approved):

eerxuan (approved):

PR #659, opened 2026-06-30
Add abs expression tests

### LINKED ISSUE #272: Add compatibility test for $abs
Part of #9.

Category: core / operator / expressions / arithmetic

Add compatibility test coverage for the `abs` feature.

Note: a smoke test exists; comprehensive compatibility test coverage is needed.

danielfrankcom:
This change adds tests for the `abs` expression. It was originally authored by @vic-tsang.

Closes #272


documentdb_tests/compatibility/tests/core/operator/expressions/arithmetic/abs/test_abs_input_forms.py
```diff
@@ -0,0 +1,68 @@
...
+        expected=5,
+        msg="$abs should return the absolute value of an inline literal argument",
+    ),
+]
+
+# Property [Expression Input]: $abs evaluates a nested expression argument before taking the
+# absolute value.
+ABS_EXPRESSION_INPUT_TESTS: list[ExpressionTestCase] = [
+    ExpressionTestCase(
+        "nested_abs",
```
PatersonProjects, line 49: Arguably unneeded case, outer $abs is a no-op

PatersonProjects (approved):
It looks good to me, but you may want to consider adding a nested field path test? 
{"a": [{"b": -1}, {"b": -2}]}, expression {"$abs": "$a.b"}

documentdb_tests/compatibility/tests/core/operator/expressions/arithmetic/abs/test_abs_input_forms.py
danielfrankcom: This is something that was retained from the original file, but I think it's defensible here since it's validating this _is_ a no-op. If the implementation did a double negation or something of the original value then this would catch it.

documentdb_tests/compatibility/tests/core/operator/expressions/arithmetic/abs/test_abs_boundaries.py
```diff
@@ -0,0 +1,252 @@
...
+        doc={"value": DOUBLE_JUST_ABOVE_HALF},
+        expression={"$abs": ["$value"]},
+        expected=DOUBLE_JUST_ABOVE_HALF,
+        msg="$abs should return the same value for a double just above half",
+    ),
+]
+
+# Property [Decimal128 Boundaries]: $abs preserves decimal128 magnitude and precision, including
+# trailing zeros.
+ABS_DECIMAL_BOUNDARY_TESTS: list[ExpressionTestCase] = [
```
ramgokul-14, line 191: [Suggestion] Per TEST_COVERAGE.md §2 (Sign handling), $abs is fundamentally a sign-flip operator, but the exponent/trailing-zero decimal128 cases here (decimal128_large_exponent, decimal128_small_exponent, decimal128_trailing_zero, decimal128_many_trailing_zeros) are all positive → positive no-ops, so they don't exercise the sign flip. decimal128_min → decimal128_max already covers the extreme magnitude. Consider adding negative counterparts (e.g. Decimal128("-1E+6144"), a negative trailing-zero value) so magnitude/precision preservation is verified through an actual negation, which is the behavior this operator is about.

documentdb_tests/compatibility/tests/core/operator/expressions/arithmetic/abs/test_abs_magnitude.py
```diff
@@ -0,0 +1,140 @@
...
+from documentdb_tests.framework.test_constants import (
+    DECIMAL128_NEGATIVE_ZERO,
+    DECIMAL128_ZERO,
+    DOUBLE_NEGATIVE_ZERO,
+    DOUBLE_ZERO,
+    INT64_ZERO,
+)
+
+# Property [Magnitude]: $abs returns the absolute value of a number, preserving its numeric type.
+ABS_MAGNITUDE_TESTS: list[ExpressionTestCase] = [
```
ramgokul-14, line 23: [Suggestion] TEST_COVERAGE.md §2 notes: "Distinguish between fractional doubles (2.5) and whole-number doubles (3.0). Some operators coerce whole-number doubles to int." The double cases here use fractional 1.5/-1.5 only. Consider adding a whole-number double case (e.g. value=-3.0, expected 3.0) to confirm $abs preserves the double type and does not coerce a whole-number double to int. A matching return_type_whole_double case in test_abs_return_type.py would make the type guarantee explicit.

danielfrankcom: I think this is already covered by the `return_type_double` test.

ramgokul-14 (approved):

eerxuan (approved):

PR #666, opened 2026-07-03
Add $arrayElemAt, $arrayToObject, and $concatArrays tests

alinaliBQ:
This change adds tests for the `$arrayElemAt`,`$arrayToObject`, and `$concatArrays` expression. They were originally authored by @leszek-bq.

Closes https://github.com/documentdb/functional-tests/issues/290, https://github.com/documentdb/functional-tests/issues/289, https://github.com/documentdb/functional-tests/issues/288


documentdb_tests/compatibility/tests/core/operator/expressions/array/utils/array_test_case.py
```diff
@@ -0,0 +1,24 @@
...
+"""
+
+from dataclasses import dataclass
+from typing import Any
+
+from documentdb_tests.framework.test_case import BaseTestCase
+
+
+@dataclass(frozen=True)
+class ArrayTestClass(BaseTestCase):
```
danielfrankcom: Is this class needed? It seems like we could use `ExpressionTestCase` for anywhere this is being used.

I get the impression `ArrayTestClass` was originally added to test both literal and field path tests with the same definition, but I think we decided after these tests were originally written that we would just test everything using field path, and then add a representative case for literals only.

We had decided to prefer writing out the entire expression rather than constructing it in the parameterized test case, as it makes it clearer how the operator is actually being invoked:

```python
ArrayTestClass(
    id="one_large_one_small",
    arrays=[list(range(10000)), [10000]],
    expected=list(range(10001)),
    msg="$concatArrays should concatenate a large array and a small array",
),
```

vs

```python
ExpressionTestCase(
    "one_large_one_small",
    doc={"arr0": list(range(10_000)), "arr1": [10_000]},
    expression={"$concatArrays": ["$arr0", "$arr1"]},
    expected=list(range(10_001)),
    msg="$concatArrays should concatenate a large array and a small array",
),
```

documentdb_tests/compatibility/tests/core/operator/expressions/array/arrayElemAt/test_arrayElemAt_core_behavior.py
```diff
@@ -0,0 +1,218 @@
+"""
```
danielfrankcom, line 1: Nit: not super important but I'm not sure why all these test files had `test_expression_<operator>` as a prefix instead of just `test_<operator>` like we have elsewhere.

I'm actually a bit surprised the framework checks allow that, I guess it's just validating the operator is _in_ the name of the test, rather than as a consistent prefix for the test.

documentdb_tests/compatibility/tests/core/operator/expressions/test_expressions_combination_array_arrayElemAt_indexOfArray_in_slice.py
```diff
@@ -0,0 +1,384 @@
...
+from documentdb_tests.compatibility.tests.core.operator.expressions.utils.utils import (
+    execute_expression,
+)
+from documentdb_tests.framework.assertions import assertResult
+from documentdb_tests.framework.parametrize import pytest_params
+from documentdb_tests.framework.test_case import BaseTestCase
+
+
+@dataclass(frozen=True)
+class CombinationTest(BaseTestCase):
```
danielfrankcom: Similar question here - can we use `ExpressionTestCase` rather than adding a dedicated class?

documentdb_tests/compatibility/tests/core/operator/expressions/test_expressions_combination_array_arrayElemAt_indexOfArray_in_slice.py
```diff
@@ -0,0 +1,384 @@
...
+def test_combination_expression(collection, test):
+    """Test array operators composed with other operators."""
+    result = execute_expression(collection, test.expr)
+    assertResult(result, expected=[{"result": test.expected}], msg=test.msg)
+
+
+# ---------------------------------------------------------------------------
+# Standalone tests for behavior that doesn't fit the dataclass pattern
+# ---------------------------------------------------------------------------
+def test_arrayElemAt_oob_is_missing_not_null(collection):
```
danielfrankcom: I think these 2 standalone tests could be collapsed into the data-driven parameterized ones:

```python
# Property [Type Via $type]: $type over $arrayElemAt reports the element's
# resolved type, including missing for out-of-bounds access.
ARRAY_ELEM_AT_TYPE_TESTS: list[ExpressionTestCase] = [
    ExpressionTestCase(
        "arrayElemAt_oob_is_missing",
        expression={"$type": {"$arrayElemAt": [[1, 2, 3], 10]}},
        expected="missing",
        msg="$arrayElemAt should produce a missing field, not null, when the index is out of bounds",
    ),
    ExpressionTestCase(
        "arrayElemAt_regex_type_preserved",
        expression={"$type": {"$arrayElemAt": [[Regex("abc")], 0]}},
        expected="regex",
        msg="$arrayElemAt should preserve the regex element type",
    ),
]
```

danielfrankcom (commented):
A few nitpicks and style questions, but these are primarily just from changes to our approach since these tests were originally written. I didn't find any coverage gaps or major issues

documentdb_tests/compatibility/tests/core/operator/expressions/array/arrayElemAt/test_arrayElemAt_core_behavior.py
alinaliBQ: 👍 fixed to have prefix `test_<operator>`

documentdb_tests/compatibility/tests/core/operator/expressions/array/utils/array_test_case.py
alinaliBQ: Good point. I have removed `ArrayTestClass` and replaced it with `ExpressionTestCase`.

documentdb_tests/compatibility/tests/core/operator/expressions/test_expressions_combination_array_arrayElemAt_indexOfArray_in_slice.py
alinaliBQ: good catch, replaced `CombinationTest` with `ExpressionTestCase`

documentdb_tests/compatibility/tests/core/operator/expressions/test_expressions_combination_array_arrayElemAt_indexOfArray_in_slice.py
alinaliBQ: 👍 I have converted these tests to use `ExpressionTestCase` with your suggestion

alinaliBQ (commented):
Addressed comments from Daniel, thanks!

danielfrankcom (approved):

danielfrankcom (approved):
Forgot to approve this after reviewing the pushed changes, LGTM

danielfrankcom (approved):

eerxuan (approved):

PR #678, opened 2026-07-10
Add $dateFromParts, $dateFromString, $dateToParts, $dateToString tests

danielfrankcom:
This change adds tests for the `$dateFromParts`, `$dateFromString`, `$dateToParts`, and `$dateToString` expressions. The tests were originally authored by @mitchell-elholm.

Closes #306, #307, #309, #310


documentdb_tests/compatibility/tests/core/operator/expressions/date/dateFromString/test_dateFromString_expressions.py
```diff
@@ -0,0 +1,156 @@
...
+        expected=datetime(2024, 6, 15, 0, 0, 0, tzinfo=timezone.utc),
+        msg="$dateFromString should accept a dateString from a nested-object path",
+    ),
+    ExpressionTestCase(
+        "nested_field_path",
+        doc={"a": {"b": "2024-06-15"}},
+        expression={"$dateFromString": {"dateString": "$a.b"}},
+        expected=datetime(2024, 6, 15, 0, 0, 0, tzinfo=timezone.utc),
+        msg="$dateFromString should accept a dateString from a distinct nested-object path",
+    ),
```
PatersonProjects: Duplicate to above

documentdb_tests/compatibility/tests/core/operator/expressions/date/dateToString/test_dateToString_types.py
```diff
@@ -0,0 +1,172 @@
...
+            ("bool_false", False),
+            ("object", {"a": 1}),
+            ("array", [1, 2]),
+            ("binary", Binary(b"\x01")),
+            ("regex", Regex(".*")),
+            ("code", Code("function(){}")),
+            ("minkey", MinKey()),
+            ("maxkey", MaxKey()),
+        ]
+    ],
```
PatersonProjects, line 55: NIT: I prefer these lists either as explicit separate ExpressionTestCases or using the BsonTypeTestCase generator, this goes for the other types files in PR as well, I think the [tid, val] setup has been used before so could be kept as it

PatersonProjects (approved):
Only thing I can spot is the one duplicate, otherwise LGTM, nice parametrizing

documentdb_tests/compatibility/tests/core/operator/expressions/date/dateToString/test_dateToString_types.py
danielfrankcom: Think I'll leave this one - we discussed something similar before but never landed on a common understanding for the best way to do this. I think there's a missing simplified model to allow this kind of basic looping but we never set it up so I've been doing something similar to this consistently so it could be easily replaced in the future.

I'm not sure I would want to define the special machinery that is necessary for a `BsonTypeTestCase` for these simple cases.

eerxuan (approved):

PR #685, opened 2026-07-14
Add $year, $month, $week, $hour tests

danielfrankcom:
This change adds tests for the `$year`, `$month`, `$week`, and `$hour` expressions. They were originally authored by @mitchell-elholm.

Closes #324, #321, #323, #315


documentdb_tests/compatibility/tests/core/operator/expressions/date/year/test_year_calendar.py
```diff
@@ -0,0 +1,231 @@
...
+    ),
+    ExpressionTestCase(
+        "ms_negative_one",
+        expression={"$year": {"$add": [DATE_EPOCH, -1]}},
+        expected=1969,
+        msg="$year should return 1969 one millisecond before the epoch",
+    ),
+    ExpressionTestCase(
+        "far_future_year_10000",
+        expression={"$year": {"$add": [DATE_EPOCH, 253_402_300_800_000]}},
```
PatersonProjects, line 174: This feels like an odd way to represent this time (goes for next 3 cases as well). Could we use datetime or a more readable constant instead?

documentdb_tests/compatibility/tests/core/operator/expressions/date/hour/test_hour_clock.py
```diff
@@ -0,0 +1,248 @@
...
+        expected=15,
+        msg="$hour should return 15 for Feb 28 in a non-leap year",
+    ),
+    ExpressionTestCase(
+        "leap_year_feb_29_2020",
+        doc={"date": datetime(2020, 2, 29, 15, 0, 0, tzinfo=timezone.utc)},
+        expression={"$hour": "$date"},
+        expected=15,
+        msg="$hour should return 15 for Feb 29 in the leap year 2020",
+    ),
```
PatersonProjects: This feels like excessive coverage for hour to me, testing with specific leap years. Also goes for month and year. Week keeping both a week 8 and week 9 leap year may be worthwhile however.

documentdb_tests/compatibility/tests/core/operator/expressions/date/year/test_year_calendar.py
```diff
@@ -0,0 +1,231 @@
...
+    ExpressionTestCase(
+        "leap_year_feb_29_2020",
+        doc={"date": datetime(2020, 2, 29, 12, 0, 0, tzinfo=timezone.utc)},
+        expression={"$year": "$date"},
+        expected=2020,
+        msg="$year should return 2020 for leap day 2020",
+    ),
+    ExpressionTestCase(
+        "leap_year_feb_29_2000",
+        doc={"date": DATE_LEAP_FEB29},
```
PatersonProjects: We actually lose the year by using the constant here, might be worth just inlining datetime.

PatersonProjects (approved):
LGTM, these are more nitpicks

documentdb_tests/compatibility/tests/core/operator/expressions/date/year/test_year_calendar.py
danielfrankcom: These four target years outside what `datetime` can represent, which is 1 through 9999. Year 10000, year 0, and the deep past and future values have no `datetime` equivalent at all, so the instant is built by offsetting from `DATE_EPOCH` in milliseconds.

The test name and expected value already carry the intent here, so I'm not sure there's a lot of value in one-off constants for this.

documentdb_tests/compatibility/tests/core/operator/expressions/date/hour/test_hour_clock.py
danielfrankcom: Good catch, I've trimmed and edited these.

For `$hour` I dropped the `2020`, `2000`, and `1900` cases and kept one `Feb 29` case. The year a leap day falls in can't change the result and the extra years were redundant. `$year` is the same story, since a leap day shifts the date by at most one day and can't cross a year boundary.

For `$month` the leap cases genuinely matter since it decodes the calendar, so I kept them but reworked them to each cover a distinct rule. I dropped the duplicate `2020` and moved `1900` to `Mar 1`.

`$week` I left as is, since the two years land on different weeks like you said.

documentdb_tests/compatibility/tests/core/operator/expressions/date/year/test_year_calendar.py
danielfrankcom: This case ended up getting removed so no longer relevant.

eerxuan (approved):

PR #684, opened 2026-07-14
Add $toInt, $toDouble, $toLong, $toDecimal tests

PatersonProjects:
This PR adds the tests for $toInt, $toDouble, $toLong, and $toDecimal. It was originally authored by @danielfrankcom. 

Ref: Issue #15, #328, #327, #329, #326

PR #688, opened 2026-07-15
Add $toString, $toBool, $toObjectId tests, and convert coverage

PatersonProjects:
This PR adds the $toString, $toBool, and $toObjectId tests. It also adds the utilities needed to run the $convert versions of the tests in this PR and in PR #684. It was originally authored by @danielfrankcom 

Ref: Issue #15, #331, #325, #330


documentdb_tests/compatibility/tests/core/operator/expressions/string/toString/test_toString_arity.py
```diff
@@ -0,0 +1,163 @@
...
+        expression={"$toString": "$$"},
+        error_code=FAILED_TO_PARSE_ERROR,
+    ),
+]
+
+
+@pytest.mark.parametrize(
+    "test", pytest_params(TOSTRING_ARITY_TESTS + TOSTRING_INVALID_FIELD_PATH_TESTS)
+)
+def test_toString_arity(collection, test: ExpressionTestCase):
```
danielfrankcom, line 158: I think every test in these files was originally parameterized over `$convert` too. These `$toX` operators are essentially an alias for `$convert: {input, to}`.

I missed this in the context of #684 but we should probably fix it there too as it's a significant coverage gap. The `$convert` tests rely on these split coverage.

documentdb_tests/compatibility/tests/core/operator/expressions/type/toBool/test_toBool_field_ref.py
```diff
@@ -0,0 +1,131 @@
...
+    ),
+    ExpressionTestCase(
+        "missing_nested_field",
+        msg="Missing nested field returns null",
+        expression={"$toBool": "$doc.missing"},
+        doc={"doc": {"x": 1}},
+        expected=None,
+    ),
+    ExpressionTestCase(
+        "composite_array_path",
```
danielfrankcom, line 86: I think we need an equivalent case for the other 2 operators, this seems to be the only one.

danielfrankcom (changes requested):
Sorry I missed this before, but all of the `$toX` operators also need to be parameterized against `$convert: {input, to}` since they are an alias, and the `$convert` coverage relies on them existing.

See [this](https://github.com/documentdb/functional-tests/blob/8dc229a79911302fe1153cfe628399effe5a56d3/documentdb_tests/compatibility/tests/core/operator/expressions/type/convert/utils/convert_common.py#L1) comment.

documentdb_tests/compatibility/tests/core/operator/expressions/string/toString/test_toString_arity.py
PatersonProjects: I have added a new type utility to allow each test to also run with $convert. It works a little differently than it originally did, so that the tests can keep using ExpressionTestCase, rather than needing their own classes, but should have the same end functionality.

documentdb_tests/compatibility/tests/core/operator/expressions/type/toBool/test_toBool_field_ref.py
PatersonProjects: Added!

danielfrankcom (approved):

eerxuan (approved):

PR #689, opened 2026-07-15
Fix autoCompact intermittent busy error

danielfrankcom:
This change fixes a flaky failure in the `autoCompact` test suite, which can be triggered by asynchronous background processing.

`autoCompact` schedules a background task that winds down asynchronously. While it's winding down, any `autoCompact` call can be transiently rejected as busy.

Previously we attempted to work around this by ensuring an idle state at the start of each run, but it is not possible to reliably detect this without delving into engine internals. To work around this, each command execution has been modified to use the `execute_admin_with_retry_command` method, which allows the tests to tolerate this background task condition. The outcomes remain the same and will be eventually consistent, we just can't reliably get the engine into a clean starting state due to lack of engine-agnostic observability.


documentdb_tests/compatibility/tests/system/administration/commands/autoCompact/utils/autoCompact_common.py
```diff
@@ -2,32 +2,23 @@
...
 from __future__ import annotations
 
-import time
-
-from documentdb_tests.framework.executor import execute_admin_command
+from documentdb_tests.framework.error_codes import OBJECT_IS_BUSY_ERROR
+from documentdb_tests.framework.executor import execute_admin_with_retry_command
 
 
 def ensure_autocompact_idle(collection):
```
eerxuan, line 9: Do we still need this when each execution can retry.

danielfrankcom: Possibly not, but I thought it seemed better to have a rough checkpoint still even if it's not 100% accurate. If 99% of tests run from a clean slate that seemed better to me than the tests regularly encountering the retry condition during their run.

I can remove it and run some load testing if you would rather remove it, just made sense to me to keep.

vic-tsang: Since we don't have permission to do re-try, I am inline with Daniel. It might be extra, but making the github action less flaky might avoid some failure investigation which was not related to the the original PR.

danielfrankcom: To be clear, in theory it should be equivalent whether we have this extra guard or not. My statement was more about where it made sense to me logically for this wait to happen. I think @eerxuan is correct that the tests should still pass reliably without this but I would need to confirm if we decide to remove it.

eerxuan: it seems they do the exact same thing. The ensure_autocompact_idle alone cannot ensure 100%, but the execution itself can. So seems the ensure_autocompact_idle is not needed.

danielfrankcom: I took a closer look at this and it needs to stay, the retry and the idle reset cover different failures.

The main job of this method is to ensure compaction is disabled, and it uses the retry to do that. There are some tests like `test_autoCompact_reconfigure_conflict` that require this as an explicit precondition, and fail if it is not disabled.

eerxuan (approved):

PR #693, opened 2026-07-17
Add stdDevPop, stdDevSamp window operator and setWindowFields stage tests

Cassie-2025:
Add compatibility tests for:
- stdDevPop window operator (11 test files, 146 tests)
- stdDevSamp window operator (11 test files, 146 tests)
- setWindowFields stage error validation (5 test files, 33 tests)

Coverage includes: documents/range/time-range frame modes, partitioning, sort order, numeric precision, non-numeric handling, NaN/Infinity behavior, field paths, pipeline stage interactions, and frame validation errors.

All 325 tests verified passing.


drk-knght (approved):
lgtm

eerxuan (commented):
Summary

  This is the first substantive window-operator PR — 325 tests across 24 files for $stdDevPop, $stdDevSamp, and $setWindowFields stage errors, where previously every window/<op>/ folder and setWindowFields/ held only smoke tests. Because it sets the template for ~26 remaining window operators, the structural decisions matter more than the individual assertions. The stdDev
  math itself is correct (population vs sample formulas, single-element pop=0 vs samp=null, deterministic sorts — no cross-contamination), but the PR bakes in significant duplication, omits the reviewer-requested TEST_COVERAGE.md spec, and deviates from the repo's near-universal test conventions.

  Findings

  1. 🔴 Window-frame / partition / sort mechanics are tested per-operator and duplicated across both operators — they belong once at the stage level.
  documents_mode_frames, range_mode_frames, time_range_mode_frames, partitioning, sort, field_paths, and pipeline_stages (≈7 of 10 files per operator) test $setWindowFields stage behavior — how documents/range/time bounds select the frame, how partitionBy/sortBy work, pipeline composition. These are spec-consistent across all ~30 window operators, and the files are
  byte-parallel between stdDevPop and stdDevSamp (identical test-name sets: whole_partition, cumulative, sliding_centered, trailing, …; verified 0 error assertions — value-only). Per TEST_COVERAGE.md §18/§19 ("test once, representative operator") and REVIEW.md §8 ("duplicate coverage"), this matrix should live once under stages/setWindowFields/ driven by one
  representative operator. Left as-is, it becomes the precedent that every future window operator re-copies the full ~48-case frame matrix ×N. Keep per-operator only the genuine stdDev delta (numeric_precision, non_numeric_handling, special_floats + 1–2 wiring cases).

  2. 🟠 docs/testing/TEST_COVERAGE.md not updated, though this introduces the window-operator feature category.
  REVIEW.md §9 (line 85): "TEST_COVERAGE.md updated if the PR introduces coverage rules for a new feature category (e.g. date operators, window functions)" — window functions is the literal example, and the PR touches the doc 0 times. As the first window-operator PR, it must add a "Window Operator Coverage" section stating what's tested once at the $setWindowFields stage
  level (frame modes, partitionBy, sortBy, frame-validation errors — representative operator) vs per-operator (the operator's own math: formula, empty/single-element result, non-numeric handling, special floats, precision). Without this written rule, finding #1's duplication becomes the standing precedent.

  3. 🟠 Test-format deviation — 322 hand-written def test_* functions instead of the parametrized-dataclass pattern.
  TEST_FORMAT.md "Parametrized Tests" and REVIEW.md §4 mandate @dataclass(frozen=True) + pytest_params() for operators with many cases. This PR uses 0 pytest_params/parametrize and 322 standalone def test_* with a copy-pasted local _run_stdDev* helper — while 1603 existing files under tests/core/operator/ use pytest_params and none use this style. This is the lone
  deviation, and it's the template the next 26 window operators would copy. (Converting to parametrized dataclasses also collapses the 800-line files, so I'm not flagging length separately.)

  4. 🟠 Shared helper + corpus duplicated per file instead of a window/utils/ module.
  Every file re-defines its own _run_stdDevPop/_run_stdDevSamp pipeline-builder and BASIC_DOCS/numeric corpus (byte-identical across the two operators bar the operator name). REVIEW.md §5/§6 and TEST_FORMAT.md require shared helpers/data in utils/; sibling accumulators/utils (AccumulatorTestCase), stages/lookup/utils, etc. establish the convention. A window-pipeline
  change would need editing ~19 times.

  5. 🟡 Raw integer error codes in the stage-error files.
  REVIEW.md §4: "Error codes from framework.error_codes used (not raw integers)." The three setWindowFields_*_errors files make 29 assertFailureCode(result, <int>) calls (code 9 twenty times, despite FAILED_TO_PARSE_ERROR = 9 existing) with 0 error_codes imports — so REVIEW.md §8 ("test the right error code") can't be verified at a glance.

  6. 🟡 Inline float("nan")/("inf")/("-inf") literals instead of the existing constants.
  TEST_COVERAGE.md:613 "All test constants … defined in framework/test_constants.py. Import from there — do not duplicate values." FLOAT_NAN/FLOAT_INFINITY/FLOAT_NEGATIVE_INFINITY already exist; the special_floats/precision files hardcode them ~80×. (Note: the files do correctly use the established assertSuccessNaN idiom for NaN comparison — that part is compliant.)

  7. 🟡 No # Property [Name]: comments in any of the 24 files.
  REVIEW.md §4 / TEST_FORMAT.md require per-group # Property [...] comments; the diff has 0. As the category template, this drops the property-documentation convention wholesale.

eerxuan (commented):
### Critical Issues

  None.

  ### Suggestions

  - [Suggestion] §7 Decimal128 boundary constants not used. numeric_precision.py covers Decimal128 via ad-hoc literals (Decimal128("10"), one 34-digit high-precision value, near-zero literals) but never touches the
  boundary matrix §7 mandates — DECIMAL128_MAX, DECIMAL128_MIN, DECIMAL128_SMALL_EXPONENT, DECIMAL128_LARGE_EXPONENT, DECIMAL128_NEGATIVE_ZERO. These constants already exist in test_constants.py. Mirror them
  instead of hardcoding, so exponent boundaries are exercised.
  - [Suggestion] Stage-level validation cases declared in §22 but untested. The §22 stage checklist the PR itself adds lists these as invalid, but no test covers them: output = empty document {} (has omitted,
  non-document, unknown-accumulator-key, but not {}); malformed sortBy spec — sortBy non-object (5), empty object {}, direction other than 1/−1 (existing sortBy error tests only cover absent sortBy or bad sortBy
  field values); window = empty document {} and window = scalar/non-document. Either add these cases or trim the §22 claims to match.
  - [Suggestion] §22 per-operator list omits the order-independence dimension. The main "Per-operator tests" bullet list enumerates 6 dimensions, but each operator folder ships 7 files — order_independence is only
  reachable via the far-down "Order-dependent vs order-independent" applicability bullet. A reader reproducing coverage for the next operator from the main list would miss it. Add an "Order independence" bullet to
  the main list.
  - [Suggestion] run_window_operator covers only the simplest shape, so ~half the files bypass it. The helper hardcodes the expression as "$value" and output field as "result", so
  field-path/argument-validation/expression tests inline full execute_command pipelines instead. An optional expression parameter (default "$value") would let those share the helper. Relatedly, the three
  near-identical _run_sum_* helpers across the stage files could collapse into one. Both are DRY cleanups, not correctness issues.

  ### Positive Feedback

  - The §22 doc addition is excellent — the "documents selected into the frame" (stage) vs. "result computed given those documents" (operator) distinction is a crisp, reusable rule, and the per-category
  applicability table (frameless / rank / N-param / gap-filler / calculus / array-output / order-dependence) will pay off well beyond this PR. Every dimension it promises for these two operators is backed by a real
  test.
  - Clean stage/operator separation, honored in code — stage files never use stdDev and never assert stddev values; operator files never touch the stage-selection matrix. No cross-duplication.
  - special_floats correctly exercises both removable and non-removable windows (whole-partition/cumulative → NaN propagation; sliding [-1,0]/[-1,1] → null), with docstrings explaining the Inf − Inf = NaN
  reasoning.
  - stdDevSamp's N−1 edge cases (single-element frame → null) are explicit in both tests and the §22 rationale.
  - Comprehensive rule adherence — honors the Global Rules (one-layer-deep, feature-folder placement, no engine internals, no error-precedence), §1 type matrix (incl. real datetime, ObjectId, regex, binary,
  minKey/maxKey), §6 (codes not messages), §8 null-vs-missing, §10 BSON distinction, negative-zero via the DOUBLE_NEGATIVE_ZERO constant, and correctly skips the non-applicable rules (§11 pipeline contexts — the
  very exclusion this PR codifies; §19/§20/§21).

eerxuan (approved):

PR #697, opened 2026-07-21
Expand collation testing for $lookup stage

danielfrankcom:
This change adds additional testing for collation in the context of the `$lookup` stage, in particular focusing on nested applications.

This is related to #673, in particular [this](https://github.com/documentdb/functional-tests/pull/673#discussion_r3585207150) comment thread.

Changes based on @PatersonProjects'.


documentdb_tests/compatibility/tests/core/collation/command_level/stages/test_stages_lookup.py
```diff
@@ -81,12 +81,12 @@ class LookupCollationTestCase(LookupTestCase):
     ),
 ]
 
-# Property [Lookup Pipeline Collation Propagation]: collation propagates into
-# the pipeline form of $lookup so that sub-pipeline stages, both $expr matches
-# and ordering, inherit command-level collation.
-COLLATION_LOOKUP_PIPELINE_TESTS: list[LookupCollationTestCase] = [
+# Property [Lookup Verbose Collation Propagation]: collation propagates into
```
yshanhu, line 84: Can we add a case: Foreign Collection Collation Ignored in Verbose correlated lookup? 
lets have:1)local collection has no collation. foreign does;2)both have collation, but different; 3)local collection has collation but foreign does not. ->in all cases, foreign collation ignored

Previously we test compatibility to see in a big picture if collation works, but we are adding more extensively for lookup due to its interaction with more collections/subpipeline.

documentdb_tests/compatibility/tests/core/collation/command_level/stages/test_stages_lookup.py
```diff
@@ -130,12 +130,197 @@ class LookupCollationTestCase(LookupTestCase):
...
+            }
+        ],
+        msg="$lookup sub-pipeline $sort without collation should order by binary comparison",
+    ),
+]
+
+# Property [Lookup Nested Sub-Pipeline Collation Propagation]: command-level
+# collation propagates through a nested $lookup into the innermost sub-pipeline,
+# so a deeply nested $expr comparison is collation-equal.
+COLLATION_LOOKUP_NESTED_TESTS: list[LookupCollationTestCase] = [
```
yshanhu, line 187: Can we have more tests for **collection levels** collation for nested lookup? focus here still: **foreign collection collation should be ignored**(please check for both verbose and concise correlated lookup as the base lookup).

Something to note:
1. In multiple level of nested lookup, can we engage **more collections** instead of mostly self joins of foreign on top of foreign collections?
2. In nested lookup, can we have interaction of **different syntax of lookup**? for example: base is verbose, but in pipeline, we can have a standard lookup nested with an uncorrelated lookup or nested with a verbose correlated lookup; Same for concise, base lookup is a concise lookup, in the pipeline, we can have combination of different forms of lookup.

documentdb_tests/compatibility/tests/core/collation/command_level/stages/test_stages_lookup.py
danielfrankcom: Added these cases for combos 1 and 2 for both verbose/concise. Combo 3 (local has collation, foreign none) is the case where the source collation still governs the foreign side despite the foreign having none, which is already covered by `lookup_inherits_source_collection_default` already.

yshanhu (approved):

eerxuan (approved):

PR #699, opened 2026-07-22
Add verbose syntax tests for $lookup

danielfrankcom:
This change adds testing for the verbose (`let` and `pipeline`) form of `$lookup`.

It focuses on how `let` variables are resolved across the sub-pipeline, including how each supported stage recognizes them, how they scope across nested `$lookup` stages, and how they remain accessible across multi-stage sub-pipelines.

The new organization structure replaces the earlier `test_lookup_with_expr.py` file.

This is related to #673 and the comments there. Changes based on @PatersonProjects' original PR.


yshanhu (commented):
Thanks for adding each stage(accessing let variable) at least one compatibility test. Looks good to me!

Just one test regarding compatibility here is, rejected stages i.e. $out, $merge should not be allowed in the lookup sub-pipeline. Could you check if we include that anywhere else in the test suite, if not, can we please add?

documentdb_tests/compatibility/tests/core/operator/stages/lookup/test_lookup_verbose_let_wiring.py
```diff
@@ -0,0 +1,512 @@
...
+    """Test $lookup sub-pipeline stages resolve let variables."""
+    with setup_lookup(collection, test_case) as foreign_name:
+        command = build_lookup_command(collection, test_case, foreign_name)
+        result = execute_command(collection, command)
+        assertResult(
+            result,
+            expected=test_case.expected,
+            error_code=test_case.error_code,
+            msg=test_case.msg,
+        )
```
yshanhu, line 512: Thanks for adding each stage(accessing let variable) at least one compatibility test. Looks good to me!

Just one test regarding compatibility here is, rejected stages i.e. $out, $merge should not be allowed in the lookup sub-pipeline. Could you check if we include that anywhere else in the test suite, if not, can we please add?

danielfrankcom: Yep, the tests for these are in `test_lookup_sub_pipeline.py` file, in the `LOOKUP_RESTRICTED_STAGES_TESTS` group.

yshanhu (approved):

eerxuan (approved):

PR #701, opened 2026-07-27
Add amazon-documentdb engine capability profile

eerxuan:
## What

Registers an `amazon-documentdb` engine target in the capability framework so the suite can run against **Amazon DocumentDB** via `--engine-name amazon-documentdb --connection-string <uri>`.

## Why

`known_engines()` currently resolves only `mongodb` and `documentdb`. Pointing the suite at an Amazon DocumentDB endpoint with `--engine-name amazon-documentdb` fails at `pytest_configure`:

```
--engine-name must be one of ['documentdb', 'mongodb'] when --connection-string is given, got 'amazon-documentdb'
```

Amazon DocumentDB is the DocumentDB engine deployed as a managed, replicated cluster. It is distinct from the open-source `documentdb` standalone profile because, being backed by a replica set, it additionally exhibits the replication-dependent capabilities the reference server gates behind one.

## Changes

- Add `("amazon-documentdb", "replica_set")` to `_CAPABILITIES_BY_PROFILE`: the DocumentDB capability surface **plus** `oplog` (the replica-set-gated capability), under a single `replica_set` topology.
- Teach `_detect_topology` to recognize `amazon-documentdb` (issues `hello` so an unreachable target resolves to no capabilities, mirroring the `documentdb` path), returning `replica_set`.

`_check_consistency()` still passes (no new capability names introduced — all already described). `known_engines()` now returns `{amazon-documentdb, documentdb, mongodb}`.

This is framework-only; no test behavior changes for existing `mongodb`/`documentdb` targets.

## Testing

- Module imports cleanly (runs `_check_consistency()` at import).
- `known_engines()` includes `amazon-documentdb`; the profile resolves to the expected capability set.
- `black`, `isort`, `flake8`, `mypy` all pass on the changed file.

### DIFF
--- a/documentdb_tests/framework/preconditions.py
+++ b/documentdb_tests/framework/preconditions.py
@@ -109,6 +109,27 @@ _CAPABILITIES_BY_PROFILE: dict[tuple[str, str], frozenset[str]] = {
             "validate_repair",
         }
     ),
+    # Amazon DocumentDB. A managed, replicated deployment: it presents the
+    # DocumentDB capability surface and, being backed by a replica set,
+    # additionally has the replication-dependent capabilities the reference
+    # server gates behind one (oplog access). Exposed as a single deployment
+    # form, so its one topology is named ``replica_set``.
+    ("amazon-documentdb", "replica_set"): frozenset(
+        {
+            "change_streams",
+            "transactions",
+            "queryable_encryption",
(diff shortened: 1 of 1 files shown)


vic-tsang (approved):
looks good to me

nitinahuja89 (approved):

PR #707, opened 2026-07-29
Add $addToSet window operator compatibility tests

akshayrao96:
## Summary

Adds per-operator MongoDB-compatibility tests for the `$addToSet` window operator (used inside `$setWindowFields`), following the existing `$stdDevPop`/`$stdDevSamp` structure and reusing the shared `window/utils/window_test_case.py` helper (`run_window_operator`, `WindowTestCase`, `BASIC_DOCS`).

## Coverage (41 new cases across 5 files)

- **`test_window_addToSet_argument_validation.py`** (12) — accepted expression forms (field path, operator, literal, array, object; each collected) + output-field/expression parse errors (unknown key, multiple accumulators, no accumulator, multi-key expression object, unknown operator, empty path component).
- **`test_window_addToSet_frame_computation.py`** (5) — documents-mode whole-partition, cumulative, reverse-cumulative, sliding (demonstrates removal as values exit the frame), and empty-frame → `[]`.
- **`test_window_addToSet_order_independence.py`** (2) — same whole-partition set regardless of `sortBy` direction.
- **`test_window_addToSet_field_paths.py`** (6) — dotted path, missing intermediate (contributes nothing), null-valued vs missing, top-level array collected whole, array-of-objects path, numeric path component.
- **`test_window_addToSet_set_semantics.py`** (16) — dedup, sliding-window removal & reappearance, numeric equivalence coalescing, bool ≠ number, null vs missing, all-missing → `[]`, document exact-duplicate collapse, document field-order distinct, array element-order distinct, mixed BSON types, string ≠ number, signed-zero collapse, ±Infinity distinct, empty `""`/`[]`/`{}`/`null`/`0` distinct, `[]` ≠ `null`, multiplicity (value persists until its last in-frame occurrence leaves), NaN collapse.

`$addToSet` output order is unspecified, so every set-valued result is compared order-independently; NaN is compared with the NaN-aware helper.

## Validation

All 42 tests (41 new + the existing smoke test) pass against **MongoDB 8.2.4**. `black`, `isort`, `flake8`, and `mypy` are clean.

## Coverage guidelines

No new coverage rule was needed — the existing Window Operator Coverage guidance (array-output operators: empty frame → `[]`, unordered output, deduplication, mixed BSON types), Numeric Equivalence in grouping/comparison, and BSON Type Distinction already cover `$addToSet`.

## Not applicable (documented)

- **`numeric_precision`** — `$addToSet` performs no arithmetic; only the value-equality slice applies, folded into set semantics (numeric coalescing).
- **standalone `special_floats`** — only NaN-as-a-set-element applies; folded into set semantics. Infinity is just another collected value (covered by the ±Infinity distinctness case).

Stage-level `$setWindowFields` mechanics (partitionBy, sortBy, range/time-range frames, default window, multiple outputs) are intentionally out of scope here — they are covered once at the stage level.

## Reference

- [`$addToSet`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/addToSet/)
- [`$setWindowFields`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/setWindowFields/)

### DIFF
--- a/documentdb_tests/compatibility/tests/core/operator/window/addToSet/test_window_addToSet_argument_validation.py
+++ b/documentdb_tests/compatibility/tests/core/operator/window/addToSet/test_window_addToSet_argument_validation.py
@@ -1,0 +1,303 @@
+"""
+Tests for $addToSet argument validation in window context.
+
+$addToSet accepts any expression as its single argument and collects the value,
+so this covers the accepted expression forms (field path, operator, literal,
+array, object) and the output-field / expression shapes that produce parse
+errors. Set-valued results are compared with ignore_order_in=["result"].
+"""
+
+from documentdb_tests.compatibility.tests.core.operator.window.utils.window_test_case import (
+    run_window_operator,
+)
+from documentdb_tests.framework.assertions import assertFailureCode, assertSuccess
+from documentdb_tests.framework.error_codes import (
+    EXPRESSION_OBJECT_MULTIPLE_FIELDS_ERROR,
+    FAILED_TO_PARSE_ERROR,
+    FIELD_PATH_EMPTY_COMPONENT_ERROR,
+    UNRECOGNIZED_EXPRESSION_ERROR,
+)
+from documentdb_tests.framework.executor import execute_command
+
+WHOLE = {"documents": ["unbounded", "unbounded"]}
+
+TWO_DOCS = [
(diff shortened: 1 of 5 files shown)


eerxuan (approved):

eerxuan (commented):
Good coverage.

Issue #117, opened 2026-04-16
Add test coverage for aggregation system variables


eerxuan:
## Summary

Add comprehensive test coverage for aggregation system variables. This includes feature tree entries, smoke tests, and detailed compatibility tests.

## System Variables to Cover

| Variable | Context | Priority |
|---|---|---|
| `$$ROOT` | Full current document | High |
| `$$CURRENT` | Current document (equivalent to field paths) | High |
| `$$REMOVE` | Remove field from output | High |
| `$$NOW` | Current datetime | High |
| `$$CLUSTER_TIME` | Current cluster time (replica set) | Medium |
| `$$DESCEND` | `$redact` stage | Medium |
| `$$PRUNE` | `$redact` stage | Medium |
| `$$KEEP` | `$redact` stage | Medium |
| `$$USER_ROLES` | Current user roles | Low |

## Tasks

- [ ] Add feature tree entries under `core,operator,expressions,variable,system-variables`
- [ ] Add smoke tests for each supported variable
- [ ] Add shared expression engine tests under `expressions/variable/system-variables/`:
  - `$$ROOT` — in `$project`, `$addFields`, nested paths (`$$ROOT.a.b`), inside `$cond`
  - `$$CURRENT` — equivalence with field paths, in various expression contexts
  - `$$REMOVE` — in `$project`, `$addFields`, conditional removal (`$cond` + `$$REMOVE`), nested fields
  - `$$NOW` — type verification, consistency within a pipeline, usage in date expressions
  - `$$DESCEND`/`$$PRUNE`/`$$KEEP` — with `$redact` stage
  - `$$CLUSTER_TIME` — type verification, availability

## Test Location

```
documentdb_tests/compatibility/tests/core/operator/expressions/variable/system-variables/
```

## Notes

- `$let` variable scoping is already covered in PR #102
- Per TEST_COVERAGE.md, system variables are shared expression engine behavior — test once with a representative operator, not per-operator

PR #710, opened 2026-07-30
Add system variable tests for $$NOW

vic-tsang:
This PR contains:
- system variable tests for $$NOW

Ref:
- Issue #117


eerxuan (commented):
- 🟡 Minor — possible over-scope on the date-operator component tests (test_now_date_operations.py::NOW_DATE_OPERATOR_TESTS). The nine $year/$month/…/$millisecond-return-int cases (plus
  $isoWeekYear→long) are really asserting each date operator's return type, which is that operator's own contract per §3 feature-ownership, not $$NOW's. The $$NOW-specific claim is "it's
  a valid date operand to the date-operator family" — a couple of representatives (one int-returning, the long-returning $isoWeekYear) prove that. The round-trip and $dateTrunc/$dateDiff
  cases are genuinely $$NOW-flavored and worth keeping. Not blocking; consider trimming the per-component matrix.

eerxuan (approved):

PR #716, opened 2026-08-07
Add $rank and $denseRank window operator compatibility tests

thanh-ttt:
## Add $rank and $denseRank window operator compatibility tests
  
  Adds comprehensive test coverage for `$rank` and `$denseRank` under `$setWindowFields`, following the same patterns as `$documentNumber`. Three test files were added for each operator (6 files total, 76 tests):
  
  | File | Dimension |
  |--------------------|-------------------------------------|
  | `test_window_*_argument_validation.py` | Asserting valid syntax (1. Of the form `{ $denseRank: { } }` or `{ $rank: { } }`, 2. Must have only 1 valid `sortBy` value, 3. No window can be specified) |
  | `test_window_*_order_dependence.py` | Ascending vs descending sort produces different rankings, ranking restarts per partition, omitted `partitionBy` treats collection as one partition |
  | `test_window_*_tie_handling.py` | No ties, all ties, partial ties at beginning/middle/end, multiple tie groups |
  
  **Key behavioral differences tested:**
  
  - `$rank`: Tied documents share the same rank, next rank **skips** positions (1, 1, 3)
     - Documentation: https://www.mongodb.com/docs/manual/reference/operator/aggregation/rank/
  - `$denseRank`: Tied documents share the same rank, next rank is **dense** / no gap (1, 1, 2)
      - Documentation: https://www.mongodb.com/docs/manual/reference/operator/aggregation/denseRank/


eerxuan (commented):
The empty-collection error tests (test_rank_window_key_errors_on_empty_collection etc.) assert the same error codes as their populated counterparts, differing only to prove the error is parse-time — a code-path distinction rather than observable behavior.

eerxuan (approved):

